PR #2, opened 2015-10-26
modification of cut

SaTahernejad:
There was a mistake with the signs of the coefficients which is fixed.


SaTahernejad:
Adding restriction on lower level variables (not complete yet)

PR #8, opened 2016-03-14
Determine instance structure, define appropriate params corresponding to instance structure and removing the error related to no good cut

SaTahernejad:


include/MibSModel.h
```diff
@@ -84,11 +84,18 @@ class MibSModel : public BlisModel {
   /** Number of structural constraints **/
   int structRowNum_;
 
+  int argnum_;
```
tkralphs: I thought we had decided there was no need to store the arguments, right? We talked about first setting defaults and then applying the user's choices afterwards.

include/MibSModel.h
```diff
@@ -84,11 +84,18 @@ class MibSModel : public BlisModel {
...
+
+  /** Determins type of problem(general or interdiction) **/
+  bool isInterdict_;
```
tkralphs: I'm not sure this is the relevant property to store. We need to know specific properties that tell us whether particular classes of cuts are valid. These properties may apply to classes other than interdiction, right? We want to have a collection of properties that are precisely the weakest properties that allow us to know whether all our classes of cuts are valid or not.

include/MibSModel.h
```diff
@@ -84,11 +84,18 @@ class MibSModel : public BlisModel {
...
+  /** Determins type of problem(general or interdiction) **/
+  bool isInterdict_; 
+  
   /** the left (negative) slope of the lower-level value function **/
   double leftSlope_;
 
   /** the right (positive) slope of the lower-level value function **/
   double rightSlope_;
+
+  std::string * arglist_;
```
tkralphs: See above comment about ```argnum_```.

include/MibSModel.h
```diff
@@ -384,6 +391,14 @@ class MibSModel : public BlisModel {
   /** The method that decodes the model from an encoded object. */
   virtual void decodeToSelf(AlpsEncoded&);
 
+  /** Determines the properties of instance. */
+  void instanceStructure(const CoinPackedMatrix *newMatrix); 
+
+  void readFromStream(std::istream& parstream);
```
tkralphs: I'm confused about the addition of this function. I don't actually see it defined anywhere... same with the others that follow.

src/MibSModel.cpp
```diff
@@ -2734,4 +2767,277 @@ MibSModel::decodeToSelf(AlpsEncoded& encoded)
...
+                    }                                                                                                                                                             
+                }                                                                                                                                                                 
+            }                                                                                                                                                                     
+        }                                                                                                                                                                         
+    }                                                                                                                                                                             
+                                                                                                                                                                                  
+    if(positiveA1){
```
tkralphs: Should we store these properties permanently somewhere so we can access them later?

include/MibSModel.h
SaTahernejad: Yes. First, I tried to write the code in the way that you explained, but I came across some issues at that tim in implementing it; Thus, I implemented that in a different way.

include/MibSModel.h
SaTahernejad: You are right. We require some other properties such as the coefficients are positive or not. I specified these properties in the function *instanceStructure* and used them. Besides, I think we still need to store the property that the problem is interdict or not because some cuts like benders cut are only valid for the interdiction problems.

tkralphs: It's not an if and only if. In one direction, Benders cuts are not valid for all interdiction problems (we must also have non-negative constraint matrix at the lower level, right?). In the other direction, Benders cuts are valid for problems that we would not necessarily label as "interdiction." At least in the current version of the code, these are problems of a very specific structure (one single budget constraint at the upper level, etc.). There is a more general class of instances for which these cuts are also valid and I'm not sure it's precisely the instances we would call "interdiction."

include/MibSModel.h
SaTahernejad: You are right. I will check them again, but I think they should be removed.

PR #14, opened 2016-06-02
Modified the bug related to no-good cut

SaTahernejad:


tkralphs:
There are actually a lot of other fixes related to this one that should be made at the same time and I have a commit already prepared with all of those fixes, so I'm going to close this pull request for now make my commit to the library branch. I would like to merge library into master soon, so we can just switch to working on that, but there's a little more work to be done first.

PR #15, opened 2016-06-02
Determine the instance structur and define the best combination of…

SaTahernejad:
… parameters

### DIFF
--- a/src/MibSCutGenerator.cpp
+++ b/src/MibSCutGenerator.cpp
@@ -428,7 +428,7 @@ MibSCutGenerator::boundCuts(BcpsConstraintPool &conPool)
 #endif
       
       NewboundModel.MibSPar()->setEntry(MibSParams::bilevelCutTypes, 1);
-      NewboundModel.MibSPar()->setEntry(MibSParams::useBendersCut, true);
+      NewboundModel.MibSPar()->setEntry(MibSParams::useBendersCut, 1);
       
       NewboundModel.MibSPar()->setEntry(MibSParams::useLowerObjHeuristic, false);
       NewboundModel.MibSPar()->setEntry(MibSParams::useObjCutHeuristic, false);
@@ -2241,7 +2241,7 @@ MibSCutGenerator::binaryCuts(BcpsConstraintPool &conPool)
   //FIXME: NEED TO CHECK FOR ROW TYPES FOR CGLP.  
   //COMING FROM KNAP SOLVER THEY ARE RANGED.
 
-  bool useNoGoodCut 
+  int useNoGoodCut 
     = localModel_->MibSPar_->entry(MibSParams::useNoGoodCut);
 
    bool useIncObjCut 
--- a/src/MibSModel.cpp
+++ b/src/MibSModel.cpp
@@ -62,7 +62,7 @@ MibSModel::MibSModel()
 //#############################################################################
 MibSModel::~MibSModel()
 {
-
+  if(arglist_) delete [] arglist_;
   if(upperColInd_) delete [] upperColInd_;
   if(lowerColInd_) delete [] lowerColInd_;
   if(upperRowInd_) delete [] upperRowInd_;
@@ -102,6 +102,16 @@ MibSModel::initialize()
   lowerRowNum_ = 0;
   upperRowNum_ = 0;
   structRowNum_ = 0;
+  isInterdict_ = false;
+  isInteger_ = true;
+  allUpperBin_ = true;
+  allLowerBin_ = true;
+  positiveA1_ = true;
+  positiveA2_ = true;
+  positiveG1_ = true;
+  positiveG2_ = true;
+  argnum_ = 0;
+  arglist_ = NULL;
   upperColInd_ = NULL;
   lowerColInd_ = NULL;
   upperRowInd_ = NULL;
@@ -137,7 +147,14 @@ MibSModel::initialize()
 void 
 MibSModel::readParameters(const int argnum, const char * const * arglist)
 {
-   //std::cout << "Reading parameters ..." << std::endl;
+
+    int i;
+    argnum_ = argnum;
+    arglist_ = new std::string[argnum];
+    for (i = 0; i < argnum; i++){
+	arglist_[i] = arglist[i];
+    }
+    
   AlpsPar_->readFromArglist(argnum, arglist);
   BlisPar_->readFromArglist(argnum, arglist);
   MibSPar_->readFromArglist(argnum, arglist);
@@ -274,6 +291,7 @@ MibSModel::readAuxiliaryData()
        m++;
      }
      else if(key == "IB"){
+	 isInterdict_ = true;
        //FIXME: ALLOW MORE THAN ONE ROW
 	data_stream >> dValue;
 	interdictBudget_ = dValue;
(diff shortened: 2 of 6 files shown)


src/MibSModel.cpp
```diff
@@ -2734,4 +2774,243 @@ MibSModel::decodeToSelf(AlpsEncoded& encoded)
...
+    }
+
+    bool paramValue(false);
+    int paramValue1(0);
+
+    //Param: "MibS_usePreprocessor"
+    paramValue = MibSPar_->entry(MibSParams::usePreprocessor);
+    if(paramValue){
+	for(i = 0; i < argnum_ - 1; i++){
+	    if(((arglist_[i] == "-param") || (arglist_[i] == "MibS_usePreprocessor"))){
```
tkralphs: Is this just because the pre-processor doesn't work? I think the error message should say something like "The pre-processor is not currently functional. Automatically disabling your parameter choice."

SaTahernejad: Hi Professor Ralphs,

You are right. I should change the error message.

Respectfully,
Sahar

src/MibSModel.cpp
```diff
@@ -2734,4 +2774,243 @@ MibSModel::decodeToSelf(AlpsEncoded& encoded)
...
+	    }
+	}
+	MibSPar()->setEntry(MibSParams::usePreprocessor, false);
+    }
+
+    //Param: "MibS_useWSHeuristic"
+    paramValue = MibSPar_->entry(MibSParams::useWSHeuristic);
+    if(paramValue){
+	for(i = 0; i < argnum_ - 1; i++){
+	    if(((arglist_[i] == "-param") || (arglist_[i] == "MibS_useWSHeuristic"))){
```
tkralphs: Does the WS Heuristic just not work? What is the reason for this?

src/MibSModel.cpp
```diff
@@ -2734,4 +2774,243 @@ MibSModel::decodeToSelf(AlpsEncoded& encoded)
...
+	    }
+	}
+	MibSPar()->setEntry(MibSParams::useWSHeuristic, false);
+    }
+
+    //Param: "MibS_useGreedyHeuristic"
+    paramValue = MibSPar_->entry(MibSParams::useGreedyHeuristic);
+    if(paramValue){
+	for(i = 0; i < argnum_ - 1; i++){
+	    if(((arglist_[i] == "-param") || (arglist_[i] == "MibS_useGreedyHeuristic"))){
```
tkralphs: Does this heuristic not work? Same question as above.

src/MibSModel.cpp
```diff
@@ -2734,4 +2774,243 @@ MibSModel::decodeToSelf(AlpsEncoded& encoded)
...
+    bool useNoGood(false);
+    
+    if((allUpperBin_) && (allLowerBin_)){
+	useNoGood = true;
+    }
+
+    paramValue = MibSPar_->entry(MibSParams::useNoGoodCut);
+
+    if(paramValue){
+	if(!useNoGood){
```
tkralphs: So here, you are saying the no good cuts are required if all variables are binary? This isn't necessarily true is it? If you have the pure integer cuts, that is technically all you need for a convergent algorithm, right? Granted, it's a good idea to have the no good cuts for binary problems, but we may want to be able to turn them off for testing purposes and if the user asks to turn the off, we have to assume they know what they're doing.

src/MibSModel.cpp
```diff
@@ -2734,4 +2774,243 @@ MibSModel::decodeToSelf(AlpsEncoded& encoded)
...
+    }
+    else{
+	if(!paramPureCut){
+	    for(i = 0; i < argnum_ - 1; i++){
+		if(((arglist_[i] == "-param") || (arglist_[i] == "MibS_useNoGoodCut"))){
+		    std::cout<<"Wrong value for MibS_useNoGoodCut. The correct value is 1."<<std::endl;
+		    assert(paramValue == true);
+		}
+	    }
+	    MibSPar()->setEntry(MibSParams::useNoGoodCut, true);
```
tkralphs: It look like the logic here is to turn on the no goods whenever the pure integer is not used, but this only works if the upper level is binary, right? Also, do we absolutely need either of them now? With the new branching scheme, we don't, right? Since we can branch on integer variables, we can have a pure branch and bound algorithm, right?

src/MibSModel.cpp
```diff
@@ -2734,4 +2774,243 @@ MibSModel::decodeToSelf(AlpsEncoded& encoded)
...
+		    assert(paramValue == false);
+		}
+	    }
+	    MibSPar()->setEntry(MibSParams::useBendersCut, false);
+	}
+    }
+    else {
+	if((paramPureCut) && (useBenders)){
+	    for(i = 0; i < argnum_ - 1; i++){
+		if(((arglist_[i] == "-param") || (arglist_[i] == "MibS_useBendersCut"))){
```
tkralphs: So the logic I think you are trying to achieve here is that if the user doesn't set the parameter value, then set it automatically. But in the case where there is a parameter file, you seem to be assuming the parameter is set. On all of these checks, this same basic logic is applied.

What I think we should be doing, rather than checking the arg list is to have a setting for each parameter called "auto". By default, these parameters would be set to the value "auto" and if the value is still "auto" when we get to this function, then we set the parameter as we see best. If the value has been changed from "auto" because the user set the parameter, then we know that the user did set the parameter specifically and we can then choose whether to over-ride or not. In this case, if the user sets the parameter, then we leave it as is. In some other cases, if the value of the parameter is set incorrectly, then we would need to over-ride. Does it make sense?

src/MibSModel.cpp
```diff
@@ -2734,4 +2774,243 @@ MibSModel::decodeToSelf(AlpsEncoded& encoded)
...
+    int * uColIndices = getUpperColInd();         
+    int * lColIndices = getLowerColInd();                                                                                                                                         
+    int * lRowIndices = getLowerRowInd();                                                                                                                                         
+                                                                                                                                                                                  
+    //Checks general or interdiction                                                                                                                                              
+    if(isInterdict_){                                                                                                                                                             
+    std::cout<<"This instance is an interdiction bilevel problem."<<std::endl;
```
tkralphs: Try to be consistent with indentation and other style elements. Of course, we can always use astyle to reformat everything.

src/MibSModel.cpp
```diff
@@ -137,7 +147,14 @@ MibSModel::initialize()
 void 
 MibSModel::readParameters(const int argnum, const char * const * arglist)
 {
-   //std::cout << "Reading parameters ..." << std::endl;
+
+    int i;
+    argnum_ = argnum;
+    arglist_ = new std::string[argnum];
+    for (i = 0; i < argnum; i++){
+	arglist_[i] = arglist[i];
```
tkralphs: I still don't think we need to store the arguments here. We should read the arguments in the below three functions and then tweak things afterward (see [this]( https://github.com/tkralphs/MibS/pull/15/commits/cb65cd686a2e7420d7446a76748f15321c6cf512#r65780141) comment below for more details).

src/MibSCutGenerator.cpp
```diff
@@ -428,7 +428,7 @@ MibSCutGenerator::boundCuts(BcpsConstraintPool &conPool)
 #endif
       
       NewboundModel.MibSPar()->setEntry(MibSParams::bilevelCutTypes, 1);
-      NewboundModel.MibSPar()->setEntry(MibSParams::useBendersCut, true);
+      NewboundModel.MibSPar()->setEntry(MibSParams::useBendersCut, 1);
```
tkralphs: Try not to use "magic numbers". To make the code readable, it's best to either use an enum or to use `#define` to define symbols to stand for the constants. So you would have something like 
```C
#define PARAM_ON
//...
NewboundModel.MibSPar()->setEntry(MibSParams::useBendersCut, PARAM_ON);
```
You can see how it's done with enums [here](https://github.com/coin-or/CHiPPS-BLIS/blob/master/Blis/src/Blis.h), for example. 
```C
enum BlisCutStrategy{
   BlisCutStrategyNotSet = -1,
   BlisCutStrategyNone = 0,
   BlisCutStrategyRoot,
   BlisCutStrategyAuto,
   BlisCutStrategyPeriodic
};
```

PR #21, opened 2016-08-23
Adding intersection cut

SaTahernejad:


include/MibSBilevel.h
```diff
@@ -47,6 +47,7 @@ class MibSBilevel {
    double *lowerSolutionOrd_;
    double *optLowerSolution_;
    double *optLowerSolutionOrd_;
+   double *lSolution_;
```
tkralphs: It looks like `lSolution_` and `optLowerSolution_` are storing the same thing here. Any reason you added `lsolution_`? In any case, try to use similar naming conventions through (spelling out `lower` versus just prepending an `l` to indicate that some data element is associated with the lower level problem).

include/MibSCutGenerator.h
```diff
@@ -70,6 +70,17 @@ class MibSCutGenerator : public BlisConGenerator {
    /** Add Benders-type cuts for zero sum problems **/
    int bendersZeroSumCuts(BcpsConstraintPool &conPool);
 
+   /** Add intersection cuts for discrete problems **/
+   int intersectionCuts(BcpsConstraintPool &conPool, double *lowerSolution);
+
+   void getAlpha(double** extRay, double* lowerSolution, int numStruct,
```
tkralphs: Maybe give this a name to make it clear that it is a helper function for intersection cuts (maybe `getAlphaIntersectionCut`). What we really should do is to create a separate C++ class for each class of cuts. Perhaps we could even use CGL directly.

include/MibSCutGenerator.h
```diff
@@ -70,6 +70,17 @@ class MibSCutGenerator : public BlisConGenerator {
...
+		 int numNonBasic, const double* lpSol,
+		 std::vector<double> &alphaVec);
+
+   double solveModel(const CoinPackedMatrix* matrix, double** extRay,
```
tkralphs: Here again (and especially in this particular case), the name should indicate that this is associated with intersection cut.

include/MibSBilevel.h
SaTahernejad: Yes, you are right. As I checked defining ` lSolution_` is useless. I will modify it.

Sure, I will change all names with` l` prefix.

include/MibSCutGenerator.h
SaTahernejad: I will modify its name. 
Yes, it seems good to have separate class for each cuts. Do you mean that I work on it now or change it whenever we start rewriting MibS again?

include/MibSCutGenerator.h
SaTahernejad: Ok, I will change it.

include/MibSCutGenerator.h
tkralphs: Just make a note of it on the global TODO list and we'll think about it later. We need to move MibS over to DisCO at some point and then we can think about re-design.

include/MibSCutGenerator.h
```diff
@@ -71,15 +71,17 @@ class MibSCutGenerator : public BlisConGenerator {
    int bendersZeroSumCuts(BcpsConstraintPool &conPool);
 
    /** Add intersection cuts for discrete problems **/
-   int intersectionCuts(BcpsConstraintPool &conPool, double *lowerSolution);
+   int intersectionCuts(BcpsConstraintPool &conPool, double *optLOowerSolution);
```
tkralphs: You have a typo here that is repeated in several other places.

src/MibSCutGenerator.cpp
```diff
@@ -562,74 +560,74 @@ MibSCutGenerator::intersectionCuts(BcpsConstraintPool &conPool, double *lowerSol
...
+MibSCutGenerator::getAlphaIntersectionCut(double** extRay, double* optLOowerSolution,
+					  int numStruct, int numNonBasic,
+					  const double* lpSol, std::vector<double> &alphaVec)
 {
     
     OsiSolverInterface * oSolver = localModel_->solver();
     
     const CoinPackedMatrix * matrix = oSolver->getMatrixByRow();
-    double * lObjCoeffs = localModel_->getLowerObjCoeffs();
+    double * lowerObjCoeffs = localModel_->getLowerObjCoeffs();
```
tkralphs: I didn't really mean that you had to change `l` to `lower` in every single instance, just that in that particular case, there were two variables declared for similar purposes right next to each other that used different conventions, so it was particularly egregious.

include/MibSCutGenerator.h
SaTahernejad: Oh, you are right. I will modify that.

src/MibSCutGenerator.cpp
SaTahernejad: I will change them again by using `l ` to be similar to the other parts of MibS.

tkralphs:
This has been merged (see 1c83ba4).

PR #29, opened 2017-02-04
MibS based on the algorithm of draft

SaTahernejad:


src/MibSBranchStrategyStrong.cpp
```diff
@@ -128,21 +144,112 @@ MibSBranchStrategyStrong::createCandBranchObjects(int numPassesLeft, double ub)
...
+			found = 1;
+			break;
+		    }
+		}
+	    }
+	}
+
+	//*********
+
+	if(branchPar == 1){
```
tkralphs: I meant to mention this to you before, we should be using `enum`'s everywhere instead of `#define`. Even if we use `#define`, this should still be `branchPar == setI`. Otherwise, why `#define setI` in the first place? `enum`'s are intended exactly for this purpose, so you can replace magic numbers with descriptions and also easily add additional options. 

In general, we want to make the code readable by anyone who knows the algorithm. Try to make the variable names reflect the actual concept they represent. If possible, don't even use names that refer to notation that may not be familiar. We should think about whether we can come up with a better name than `setI`, since that only makes sense in the context of that one paper and we could change that notation anytime.

src/MibSBranchStrategyStrong.cpp
```diff
@@ -128,21 +144,112 @@ MibSBranchStrategyStrong::createCandBranchObjects(int numPassesLeft, double ub)
...
+		else{
+		    candidate[i] = 2;
+		    value =saveSolution[i];
+		    infeasibility =fabs(floor(value + 0.5) - value);
+		    if(infeasibility > etol){
+			candidate[i] = 1;
+		    }
+		}
+	    }
+	}
```
tkralphs: These loops are almost identical. You should be able to compress them into one without losing too much effficiency (if any). It's best to avoid duplication of identical lines of code.

src/MibSConstants.hpp
```diff
@@ -29,8 +29,13 @@
 #define PARAM_ON       1
 
 #define PARAM_NOTSET  -1
-#define notRestricted  0
-#define Restricted     1
+#define fractional     0
+#define setI           1
```
tkralphs: These should be `enum`'s rather than `#define`, as mentioned earlier.

src/MibSConstants.hpp
```diff
@@ -29,8 +29,13 @@
...
+
+/*---------  status of bilevel feasibility (bilevelFeasibility_) ------------*/
+#define unknown -1
+#define bilevelInfeasible 0
+#define bilevelFeasible 1
```
tkralphs: Here, too.

src/MibSBranchStrategyStrong.cpp
SaTahernejad: Sorry! I missed these parts. 

Sure, I will use enum.

I think maybe "restrictedToSetI" and "restrictedToFractional" are better options. What do you think?

src/MibSBranchStrategyStrong.cpp
SaTahernejad: Sure.

src/MibSBilevel.cpp
```diff
@@ -117,43 +137,114 @@ MibSBilevel::createBilevel(CoinPackedVector* sol,
...
+  int numSavedSol(sizeSetE/(sizeFixedInd+1));
+  int solType(0);
+  int num(0);
+  bool found(true);
+
+  if(setEPar != PARAM_ON){
+      isContainedInSetE_ = false;
+  }
+  else{
+      if(isIVarsIntegral_ == true){
```
tkralphs: Here, you can just say `If (isIVarsIntegral_){...`. The reason to put the `is` is in front is so that this then reads like a sentence. There are some style things we should talk about, but those can be fixed by running the code through a formatter. Personally, I like a space after `if` and the `else` on the same line as `}`.

src/MibSConstants.hpp
```diff
@@ -28,9 +28,21 @@
 #define PARAM_OFF      0
 #define PARAM_ON       1
 
+//#############################################################################    
+
+enum MibSBranchingProcedure{
+    MibSBranchingProcedureNotSet = -1,
+    fractional,
+    setI
```
tkralphs: You're kind of mixing two naming conventions here. The first entry is prefixed by `MibSBranchingProcedure` and the second and third aren't. If you follow the convention of BLIS (see Blis.h), then all the entries in the enum should be prefixed by `MibSBranchingProcedure`. Also, BLIS calls it `BranchingStrategy` so better to use the same terminology.

src/MibSBranchStrategyStrong.cpp
SaTahernejad: It is fixed.

src/MibSConstants.hpp
SaTahernejad: It is fixed.

src/MibSConstants.hpp
SaTahernejad: It is fixed.

src/MibSBilevel.cpp
```diff
@@ -47,13 +47,30 @@ MibSBilevel::createBilevel(CoinPackedVector* sol,
...
   double etol(model_->BlisPar()->entry(BlisParams::integerTol));
-  //double etol(model_->etol_); 
+
+  
+
+  MibSBranchingStrategy branchPar = static_cast<MibSBranchingStrategy>
+      (model_->MibSPar_->entry(MibSParams::branchStrategy));
+
+  int solveLowerXYVarsInt(model_->MibSPar_->entry
+			   (MibSParams::solveLowerWhenXYVarsInt));
```
tkralphs: For these parameters, I think it's clearer if we name them as in the paper: the parameter name itself should be something like `solveLowerWhenXYVarsInt` and the value is just `True` or `False`. Then you can have statements like
```
if (solveLowerWhenXYVarsInt){
...
}
```
which read logically like a sentence, just as in the paper. There's no need for an additional enum if the parameter only has values essentially equal to "on" and "off".

src/MibSBilevel.cpp
```diff
@@ -47,13 +47,30 @@ MibSBilevel::createBilevel(CoinPackedVector* sol,
...
+  int solveLowerXVarsInt(model_->MibSPar_->entry
+			  (MibSParams::solveLowerWhenXVarsInt));
+  int solveLowerIVarsInt(model_->MibSPar_->entry
+			  (MibSParams::solveLowerWhenIVarsInt));
```
tkralphs: Weren't we going to change `IVars` to `LVars` or even `LinkVars`?

src/MibSBilevel.cpp
```diff
@@ -1031,3 +1184,106 @@ MibSBilevel::getLowerObj(const double * sol, double objSense)
    return objVal * objSense;
 
 }
+//#############################################################################
+void
+    MibSBilevel::addSolutionToSetE(MibSSetETag solTag, std::vector<double>
+				   &shouldStoreValues, double objValue)
```
tkralphs: Let's come up with a different way of referring to "set E" that will make it clear what it actually is without having to refer to the paper. We should also refer to it the same way in the paper I suppose.

src/MibSBilevel.cpp
```diff
@@ -1031,3 +1184,106 @@ MibSBilevel::getLowerObj(const double * sol, double objSense)
...
+	    value = (int)(upperSolutionOrd_[i] + 0.5);
+	    model_->setE_.push_back(value);
+	}
+    }
+    model_->setE_.push_back(solType);
+    indexInSetE_ = (model_->setE_.size()/(sizeFixedInd + 1));
+    isContainedInSetE_ = true;
+
+    switch(solTag){
+    case MibSSetETagVFIsInfeasible:
```
tkralphs: "SetETag" is also not very descriptive.

src/MibSBilevel.cpp
```diff
@@ -1031,3 +1184,106 @@ MibSBilevel::getLowerObj(const double * sol, double objSense)
...
+	{
+	    solTagInSetE_ = MibSSetETagVFIsInfeasible;
+	    model_->addressInSolPool_.push_back
```
tkralphs: What does `addressInSolPool` refer to?

src/MibSBilevel.cpp
```diff
@@ -1031,3 +1184,106 @@ MibSBilevel::getLowerObj(const double * sol, double objSense)
...
+	}
+    }
+    model_->setE_.push_back(solType);
+    indexInSetE_ = (model_->setE_.size()/(sizeFixedInd + 1));
+    isContainedInSetE_ = true;
+
+    switch(solTag){
+    case MibSSetETagVFIsInfeasible:
+	{
+	    solTagInSetE_ = MibSSetETagVFIsInfeasible;
```
tkralphs: What is the purpose of storing this tag in a variable at class scope? It must be temporary storage for some reason? This is usually asking for trouble because it's not clear when the information is "stale" or what it refers to if debugging some other part of the code. Once I read more of the code, I'm sure I'll understand your concept better, but although there are already a lot of such variables used in MibS, this is not really a coding practice to be emulated :).

src/MibSConstants.hpp
tkralphs: By what commit? It still seems to be this way in the current diff for the pull request...

tkralphs: Oh, I see, you commented it out. Probably better to delete...

src/MibSModel.cpp
```diff
@@ -3209,6 +3186,58 @@ MibSModel::instanceStructure(const CoinPackedMatrix *newMatrix, const double* ro
 	}
     }
 
+    if(MibSPar_->entry(MibSParams::branchStrategy) ==
+       MibSBranchingStrategyLinking){
+	std::cout << "Branching procedure is set to 'MibSBranchingStrategyLinking' (new)."
```
tkralphs: Let's not use "new" and "old" :).

src/MibSConstants.hpp
SaTahernejad: Sure, I will delete it.

src/MibSBilevel.cpp
SaTahernejad: I am a little confused because I have not defined enum for these parameters :). Moreover, I named these parameter as in the paper, but I cannot store the value of parameter in a variable with the same name. Do you agree with me?

src/MibSBilevel.cpp
SaTahernejad: You are right. I will modify it.

src/MibSBilevel.cpp
SaTahernejad: I will think about it.

SaTahernejad: What do you think about set "seenLinkVars"?

src/MibSBilevel.cpp
SaTahernejad: Let me explain the story of storing the linking variables and the corresponding solutions of (VF) and (UB) briefly.

After solving problem (VF) corresponding to `x_{\J}=x^t_{\J}`, we add `(x^t_{\J},MibSSetETagVFIsInfeasible)` or `(x^t_{\J},MibSSetETagVFIsFeasible)` to the end of "`setE_`" vector based on the feasibility of problem (VF). In the case of feasibility of this problem, we will add the optimal solution and the optimal objective value of (VF) to the end of "`solutionPoolSetE_`" vector.  Moreover, we will add the index of the starting point of adding these solutions to  "`solutionPoolSetE_`" and the length of added values to this vector (which is `n_2 + 1` for this case) respectively to the end of "`addressInSolPool_`" vector. In the case of infeasibility of (VF), we do not add anything to "`solutionPoolSetE_`" and add the current size of "`solutionPoolSetE_`" and 0 to the end of "`addressInSolPool_`" vector. 

In the case of solving problem (UB) corresponding to `x_{\J}=x^t_{\J}`, since problem (VF) was solved before corresponding to these linking variables, `(x^t_{\J},MibSSetETagVFIsFeasible)`  has been stored before in "`setE_"` vector, so first of all we remove `(x^t_{\J},MibSSetETagVFIsFeasible)` from `setE_` and add `(x^t_{\J},MibSSetETagUBIsSolved)` to the end of this vector. In the case of infeasibility of (UB), we remove the stored optimal solution and objective value of (VF), which were stored before in "`solutionPoolSetE_`" vector and add these values to the end of this vector. Since we have removed one piece of "`solutionPoolSetE_`" vector from the middle of this vector, we update the stored values in "addressInSolPool_" vector and add the index of starting point of adding the optimal solution of (VF) and `(n_2+1)` respectively to the end of "addressInSolPool_" vector. 
In the case of feasibility of (UB), after removing the stored optimal solution and objective value of (VF) from "`solutionPoolSetE_`" vector, we add the optimal solution of (UB), the optimal objective value of (VF) and optimal objective value of (UB) respectively to the end of this vector. After updating "`addressInSolPool_`" vector, we add the index of starting point of adding the optimal solution of (UB) and `(n_1+n_2+2)` respectively to the end of "addressInSolPool_" vector.

src/MibSBilevel.cpp
SaTahernejad: I have defined it in this way because we also need solTag in `MibSBilevel` and `MibSCutGeneration` classes.

src/MibSModel.cpp
SaTahernejad: Sure :).

src/MibSBilevel.cpp
tkralphs: OK, yes, I jumped too quickly to conclusions. I believe you can have a local variable with the same name, since the scope of the other variable is the `MibSParams` class, so there would be on conflict. You could also just figure out whether to solve the lower level problem using the parameters directly (no local variables to store parameter values, if you know what I mean) and just store *that* true/false value for later use. But that would make the logic look a little different than that is in the paper, whereas it currently matches up nicely. Does it make sense?

src/MibSBilevel.cpp
tkralphs: It really is a sort of solution pool, just like the usual solution pool, so I think we should use the "solution" somehow in the name. What about `seenSolutions`? I know they aren't really full solutions, though. We need a word for a vector of linking variables that is a subset of a full first-level solution. How about about `seenLinkingSolutions` or `seenLinkingVectors`?

src/MibSBilevel.cpp
tkralphs: This sounds a  bit overly complicated :). I still have to fully absorb it and look at the code itself in detail, but it seems that set E should be a single data structure containing all linking solutions we've seen so far, with their corresponding tags, second-level solutions, etc. I would probably define a simple structure to contain all of these (but it might also need to be a class):
```
struct LINKING_SOLUTION{
int tag;
double secondLevelObjVal;
...
} linkingSolution;
```
I would store these in a hash table (Map) with the key being the solution itself , since you need to be able to easily determine if a given solution is already in the pool. When the tag of a solution already in the pool changes, we just update the tag. There's no extra bookkeeping this way and lookups are efficient. Of course, there could be even better ways, this is just off the top of my head. Does it make sense?

src/MibSBilevel.cpp
SaTahernejad: I am not sure if I have understood your second suggestion correctly. Do you mean after determining the values of "isIntegral_", "isUpperIntegral_", "isLinkVarsInt_" and "isLinkVarsFixed_" variables, we define a new boolean variable like "shouldSolveLower" and specify its value based on the values of parameters without storing them in local variables. Do you mean that?

src/MibSBilevel.cpp
SaTahernejad: I think `seenLinkingSolutions` is the best suggestion. Do you agree with me?

src/MibSBilevel.cpp
tkralphs: OK, but this isn't a very transparent way to pass data from one component to another. We should at least call it something like `mostRecentSolTag` so it's clear exactly what it is. If this is used to figure out whether we are forced to branch or forced to cut, then another possibility would be just to have a flag that indicates *that* instead and figure out whether that flag should be set right after solving (UB) or (VF). Or a third option is to just design a better mechanism for passing information between the different components of the algorithm. I suppose we could just put this on the global TODO list for the re-factoring we're going to undertake next :).

src/MibSBilevel.cpp
tkralphs: Yes, right, that was roughly the idea.

src/MibSBilevel.cpp
tkralphs: Sure.

src/MibSBilevel.cpp
SaTahernejad: I know my idea is very complicated:).
I am studying hash tables in C++ now, but I think I have understood what you mean. Let me ask possible questions after thinking more about it.

src/MibSBilevel.cpp
SaTahernejad: Sure. I will implement one of your suggestions and add the last suggestion to TODO list:).

src/MibSBilevel.cpp
SaTahernejad: I like more your first idea ,i.e. using local variables with the same names as parameters because it is clearer. Do you agree with me?

src/MibSBilevel.cpp
SaTahernejad: I will implement it in both of the code and the draft.

src/MibSBilevel.cpp
tkralphs: Sure, whatever you think :)

src/MibSBilevel.cpp
SaTahernejad: As I understood, you mean to define `LINKING_SOLUTION` structure and `std::map<std::vector<double>, LINKING_SOLUTION> seenLinkingSolutions` as members of MibSModel class, then after observing each linking solution, we store its tag, second-level objective value and .... to `linkingSolution` and add it to `seenLinkingSolutions` with x_{\J} as the key. Please let me know if I have missed any parts.

src/MibSCutGenerator.cpp
```diff
@@ -1255,7 +1255,11 @@ MibSCutGenerator::boundCuts(BcpsConstraintPool &conPool)
       
       int argc = 1;
       char** argv = new char* [1];
-      argv[0] = "mibs";
+      //argv[0] = "mibs";
+      argv[0][0] = 'm';
+      argv[0][1] = 'i';
+      argv[0][2] = 'b';
+      argv[0][3] = 's';
```
tkralphs: Hmm, this is a little convoluted :). I'm not even sure why this works, since you are never allocating space for the string, just adding to it. I guess maybe something like 
```
argv[0] = (char *) mibs
```
should work. If you Google "ISO C++ forbids converting a string constant to 'char*'", there are lots of solutions out there, e.g., http://stackoverflow.com/questions/1524356/c-deprecated-conversion-from-string-constant-to-char

SaTahernejad: I see what you mean:). I fixed it.

src/MibSBilevel.cpp
SaTahernejad: The names of variables were changed, as you suggested.

src/MibSHelp.hpp
```diff
@@ -0,0 +1,29 @@
...
+
+struct LINKING_SOLUTION{
+    int tag;
+    double lowerObjValue;
+    double UBObjValue;
+    std::vector<double> lowerSolution;
+    std::vector<double> UBSolution;
+};
+
+#endif
```
tkralphs: Not sure we need a new file just for this. It should probably go into one of the other headers, like maybe MibSSolTypes.hpp :)? Actually, I didn't evven know that header file was there, looks like it's used for data structures to store solutions for heuristics, maybe.

Just for future reference, I guess that you meant something like `MibSHelper.hpp`? `MibSHelp.hpp` seems to indicate it's about the help message for the user or something (at least to me). 

If we want to leave it in a separate header for different global types, I would call it something like `MibSTypes.hpp`.

tkralphs:
Merging, wahoo!

PR #33, opened 2017-04-21
finalMibSPaper

SaTahernejad:
This is the last version used for running the tests.

### DIFF
--- a/src/MibSBilevel.cpp
+++ b/src/MibSBilevel.cpp
@@ -167,28 +167,25 @@ MibSBilevel::createBilevel(CoinPackedVector* sol,
 
   int pos(0);
 
-  for(i = 0; i < numElements; i++){
+  for (i = 0; i < numElements; i++){
       index = indices[i];
       pos = binarySearch(0, lN - 1, index, lowerColInd);
-      if(pos < 0){
+      if (pos < 0){
 	  pos = binarySearch(0, uN - 1, index, upperColInd);
-	  if((mibs->solver()->isInteger(index)) &&
-	     (((values[i] - floor(values[i]))<etol) ||
-	      ((ceil(values[i]) - values[i]) < etol))){
+	  if ((mibs->solver()->isInteger(index)) &&
+	      (((values[i] - floor(values[i])) < etol) ||
+	       ((ceil(values[i]) - values[i]) < etol))){
 	      upperSolutionOrd_[pos] = (double) floor(values[i] + 0.5);
+	  }else{
+	      upperSolutionOrd_[pos] = values[i];
 	  }
-	  else{
-		  upperSolutionOrd_[pos] = values[i];
-	      }
 	  optUpperSolutionOrd_[pos] = upperSolutionOrd_[pos];
-      }
-      else{
-	  if((mibs->solver()->isInteger(index)) &&
-	     (((values[i] - floor(values[i]))<etol) ||
-	      ((ceil(values[i]) - values[i]) < etol))){
-		  lowerSolutionOrd_[pos] = (double) floor(values[i] + 0.5);
-	  }
-	  else{
+      }else{
+	  if ((mibs->solver()->isInteger(index)) &&
+	      (((values[i] - floor(values[i])) < etol) ||
+	       ((ceil(values[i]) - values[i]) < etol))){
+	      lowerSolutionOrd_[pos] = (double) floor(values[i] + 0.5);
+	  }else{
 	      lowerSolutionOrd_[pos] = values[i];
 	  }
 	  optLowerSolutionOrd_[pos] = lowerSolutionOrd_[pos];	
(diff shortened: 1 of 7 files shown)


src/MibSBilevel.cpp
```diff
@@ -171,22 +171,12 @@ MibSBilevel::createBilevel(CoinPackedVector* sol,
...
+	upperSolutionOrd_[pos] = values[i];
+	optUpperSolutionOrd_[pos] = upperSolutionOrd_[pos];
     }
     else{
-	if(mibs->solver()->isInteger(index)){
-	    lowerSolutionOrd_[pos] = (double) floor(values[i] + 0.5);
-	}
-	else{
-	    lowerSolutionOrd_[pos] = values[i];
-	}
```
tkralphs: Hmm ,what is the reason for this?

src/MibSBilevel.cpp
```diff
@@ -511,21 +564,11 @@ MibSBilevel::checkBilevelFeasiblity(bool isRoot)
 			shouldStoreValues.push_back(valuesUB[i]);
 			pos = binarySearch(0, uN - 1, i, upperColInd);
 			if (pos >= 0){
-			    if(UBSolver->isInteger(i)){
-				optUpperSolutionOrd_[pos] = (double) floor(valuesUB[i] + 0.5);
-			    }
-			    else{
-				optUpperSolutionOrd_[pos] = (double) valuesUB[i];
-			    }
+			    optUpperSolutionOrd_[pos] = (double) valuesUB[i];
```
tkralphs: Same question here.

src/MibSCutGenerator.cpp
```diff
@@ -830,7 +847,67 @@ MibSCutGenerator::storeBestSolIntersectionCutType2(const double* lpSol,
...
+			      DO_NOT_GENERATE);
+	    sym_set_int_param(env, "generate_cgl_probing_cuts",
+			      DO_NOT_GENERATE);
+	    sym_set_int_param(env, "generate_cgl_clique_cuts",
+			      DO_NOT_GENERATE);
+	    sym_set_int_param(env, "generate_cgl_twomir_cuts",
+			      DO_NOT_GENERATE);
+	    sym_set_int_param(env, "generate_cgl_flowcover_cuts",
+			      DO_NOT_GENERATE);
+	}
```
tkralphs: We should try to avoid having big repeated chunks of code. We can make a simple helper function so that it we want to make a change, we don't have to do it in multiple places.

src/MibSBilevel.cpp
SaTahernejad: I wanted to use `optLowerSolutionOrd_` here
https://github.com/SaTahernejad/MibS/blob/finalMibSPaper/src/MibSModel.cpp#L1720-L1722

as the optimal solution of relaxation problem.

src/MibSCutGenerator.cpp
SaTahernejad: You are right. I will modify these parts later.

src/MibSBilevel.cpp
tkralphs: I don't really understand. All that seems to be done here is that the code that rounds the solution to be an exact integer value has been removed, which will cause problems. I was asking why that was done.

SaTahernejad: But the optimal solution of relaxation problem may not satisfy integrality constraints.

src/MibSBilevel.cpp
SaTahernejad: The same with above answer.

src/MibSBilevel.cpp
tkralphs: The code that you deleted was there before for some reason. Are you saying it was a bug? If not, then what is the reason why it is valid to remove it now when before it was not. I'm just not seeing why you did that. It's not just in one place, it's in lots of places. I'm pretty sure I remember adding that code for a good reason, because it fixed some other bug.

src/MibSBilevel.cpp
tkralphs: Same additional question :).

src/MibSBilevel.cpp
SaTahernejad: Let x1 = 1.2 and x1 is an integer variable, then by using the previous code, we have 

`upperSolutionOrd_[1] = double(floor(1.2+0.5)) = 1.00`

Hence, `upperSolutionOrd_[1]` does not save the correct value of `x1`, when it does not satisfy integrality requirements. It was not a bug before because we used `upperSolutionOrd_` , `lowerSolutionOrd_`, `optUpperSolutionOrd_` and `optLowerSolutionOrd_` only for the situations in which all of the integrality requirements were satisfied. I thought it may be better to modify it :). I think the reason that the code was written in this way before is for avoiding the numerical errors, so when x1=1.0000001, we will have `upperSolutionOrd_[1] = 1.000`. My suggestion is to use the previous idea for rounding the values of integer variables to the closest integer value only when integer variables satisfy integrality requirements. Do you agree with me?

src/MibSBilevel.cpp
SaTahernejad: The same with above answer:).

src/MibSBilevel.cpp
tkralphs: Yes agreed.

SaTahernejad: It has been done.

src/MibSBilevel.cpp
```diff
@@ -168,17 +168,31 @@ MibSBilevel::createBilevel(CoinPackedVector* sol,
...
+      index = indices[i];
+      pos = binarySearch(0, lN - 1, index, lowerColInd);
+      if(pos < 0){
+	  pos = binarySearch(0, uN - 1, index, upperColInd);
+	  if((mibs->solver()->isInteger(index)) &&
+	     (((values[i] - floor(values[i]))<etol) ||
+	      ((ceil(values[i]) - values[i]) < etol))){
+	      upperSolutionOrd_[pos] = (double) floor(values[i] + 0.5);
+	  }
+	  else{
```
tkralphs: Can you put your `else`'s on the same line as the closing bracket of the `if` and also put a space after `for` and `if` :). We should just run the code through an automatic  formatter at some point.

src/MibSBilevel.cpp
```diff
@@ -168,17 +168,31 @@ MibSBilevel::createBilevel(CoinPackedVector* sol,
...
+		  upperSolutionOrd_[pos] = values[i];
+	      }
+	  optUpperSolutionOrd_[pos] = upperSolutionOrd_[pos];
+      }
+      else{
+	  if((mibs->solver()->isInteger(index)) &&
+	     (((values[i] - floor(values[i]))<etol) ||
```
tkralphs: Also put spaces around `<` :)? Do we want to just check whether the whole solution is integer rather than each individual component? That would make the coe a bit cleaner.

src/MibSBilevel.cpp
SaTahernejad: I implemented your advices for the new added parts, however, as you said, we should use a code formatting tool  to spruce MibS up. Do you have any suggestions for the code formatter?

src/MibSBilevel.cpp
SaTahernejad: I think it is better to check the components separately, so we round the components which satisfy integrality requirements to the closest integer value.

PR #34, opened 2017-04-22
Name based input

SaTahernejad:
The version which allows the names for variables and constraints.


tkralphs:
All the commits up until 61f0c97 are the same as those in #33. There is only a small conflict, which could easily be resolved, but I guess that it's still better to branch from current master and cherry pick the two new commits.

tkralphs:
@SaTahernejad  can you make a new pull request with just the last two commits?

SaTahernejad:
Sure.

PR #35, opened 2017-05-04
Adding the possibility of using names for variables and constraints

SaTahernejad:


tkralphs:
Looks like everything is OK, but can you maybe add something to the testing that tests this capability? If you don't know exactly how to go about that, can you just add a model that uses this capability? I guess you must have constructed one for your own testing.

SaTahernejad:
Sure, I have tested the code with the Moore instance in which the constraints and variables have names. I had tested the code by using the parameter file before, but without it I get error. I will add the instance after debugging the code.

SaTahernejad:
I added the instance in "data" file. Just note that for running this instance, the "MibS_inputFormat" parameter should be set to "nameBased". Its default value is "indexBased".

data/moore90WithName.txt
```diff
@@ -0,0 +1,9 @@
+N 1
+M 4
+LC LV
+LR R1
+LR R2
+LR R3
+LR R4
+LO 1
```
tkralphs: I guess you didn't make it possible to specify the objective by name?

SaTahernejad: Let me check it.

SaTahernejad: One technical question :). Where is important the name of objective?

tkralphs: I didn't quite get your question. I didn't really mean the name of the objective, but the names of the variables in specifying the objective. It's the same issue as with the constraints. Right now, the format is fragile because it assumes that everything is indexed inside the solver the same way it is in the original instance file. This applies to variables as well as constraints. The `LO` rows are lower objective coefficients and we should be able to specify these also by name.

SaTahernejad: We can use the name of lower level variables instead of `LO`. For example :  `LV 1`
Do you agree with it?

tkralphs: Now that I think more about it, you can probably just put the objective coefficient on the same line as the variable name, something like
```
LC LV 1
```
There doesn't seem to be much reason for another line for the same variable, right? As a matter of fact, we don't even technically need the markers `LC` and `LR`, since rows and variables must have unique names (I suppose?). We could have a format in which we have sections, one for lower constraints, one for lower variables and they are all just listed. It would be OK to have multiple supported formats, this is the case for the file formats for decompositions, there are a few different ones.

SaTahernejad: I agree with your idea about having something like `LC LV 1`. As you explained, it is more efficient than removing `LC` and `LR`. I will work on implementing this idea.

I will also work on having different sections for lower variables and constraints.

tkralphs: But I also explained that you can avoid the `LC` and `LR` labels _and_ also have efficiency with sections. Did you get that?

SaTahernejad: So you mean to have two formats:

1. We remove `LC` and `LR` completely, so we will have something like `LV 1` and remove `LC 1`.

2. We will have two sections in .txt format. I think we need keys for separating the sections.    For example, we can use `end of Vars` for  variable section and `end of Cons` for constraints.

Have I understood correctly what you mean?

tkralphs: Yes, two formats. If 1 and 2 above were meant only to describe two aspects of the second format and not both formats, then I think we are on the same page. Making the separating keys simple text strings (no spaces) will make it easier to parse. I guess it would be best to avoid any strings that could be possible variable names. Also, we should support the addition of comment lines, so you can specify a comment character that beings any line that should be skipped. Using `#` would maintain consistency with Python and Bash comments.

SaTahernejad: OK, I understood the second format.

I thought the first format is the one, which is explained in 1, i.e, it is similar to the second format without sections for separating the variables and constraints. Is it correct?

tkralphs: The first format is the one we already have. The second format is the one with sections. I guess the format without sections and without markers like `LC` and `LR` isn't that useful.

SaTahernejad: Thanks! I will implement the format with sections.

SaTahernejad: The new format has been implemented.

tkralphs:
I made a new PR #43 (I couldn't push to your pull request for some reason) deleting 55d574b, which was superfluous.

PR #38, opened 2017-09-22
Turning off the heuristics in default setting

SaTahernejad:


tkralphs:
Are the heuristics broken or are there just assumptions made that may not be met by all instances?

SaTahernejad:
No

PR #44, opened 2017-11-06
Some modifications in scripts

SaTahernejad:


scripts/README.md
```diff
@@ -0,0 +1,71 @@
+# Building and Installing MibS
```
tkralphs: I guess this is covered in the man README and we don't want to maintain two different versions of the build instructions, so you can just refer to the main README for building. Here, just explain that the purpose of the scripts is and how to use them. Also, in the main README, you could add a sentence or two pointing out that there are scripts for running tests in the `scripts` directory.

PR #46, opened 2018-01-04
Fix for #45

### LINKED ISSUE #45: Infeasibility not found
We are running the latest version of MibS on Windows 10 to model a bilevel optimization problem in the field of auction design. 

It would be great if you could help us with a problem which came up. In our model, lower level variables appear in upper level constraints and an optimal solution of the lower level can create infeasibility in the upper level. We are not sure whether this problem conflicts with the definition of bilevel programs or with the models you allow in your algorithm.

Consider the following model as a simple example:
min -x
s.t. y <= 0 
     y \in argmax {y, s.t. y <= 2, y \in R}
     5 >= x >= 0; x \in Z

Here, the only solution which fulfills the second constraint (maximizing y with the single constraint y <= 2) is y = 2, which in turn leads to infeasibility in the upper level in combination with the first constraint (y <= 0). Therefore, as we understand it, the full model is **infeasible**. Of course, we can make the model more sophisticated, allowing for the lower level variable to not appear alone in an upper level constraint. However, using MibS on this model we get a **‘feasible’** solution with x = 5; y = 0:

We are sorry if we are missing something very fundamental here, like that the upper level constraints need to be respected in the lower level program – however to us the definitions did not force this.
We would be very grateful if you could help us with this problem. Please find attached the input files we use and the complete log.

Best regards,
Richard Littmann

[infeasAux.txt](https://github.com/tkralphs/MibS/files/1597488/infeasAux.txt)

[infeasMPS.mps.txt](https://github.com/tkralphs/MibS/files/1597489/infeasMPS.mps.txt)

[infeasLog.txt](https://github.com/tkralphs/MibS/files/1597498/infeasLog.txt)

SaTahernejad:
This fixes #45.

### DIFF
--- a/src/MibSModel.cpp
+++ b/src/MibSModel.cpp
@@ -1837,7 +1837,7 @@ MibSModel::userFeasibleSolution(const double * solution, bool &userFeasible)
   if(bS_->shouldPrune_){
       userFeasible = true;
       if(!bS_->isProvenOptimal_){
-	  upperObj = 10000000;
+	  upperObj = 1e+20;
       }
   }
 
--- a/src/MibSModel.cpp
+++ b/src/MibSModel.cpp
@@ -1836,12 +1836,12 @@ MibSModel::userFeasibleSolution(const double * solution, bool &userFeasible)
   userFeasible = false;
   if(bS_->shouldPrune_){
       userFeasible = true;
-      if(!bS_->isProvenOptimal_){
-	  upperObj = 1e+20;
-      }
   }
 
-  if(userFeasible == true){
+  if(!bS_->isProvenOptimal_){
+      userFeasible = false;
+  }
+  else if(userFeasible == true){
       mibSol = new MibSSolution(getNumCols(),
 				lpSolution,
 				upperObj,
--- a/src/MibSModel.cpp
+++ b/src/MibSModel.cpp
@@ -1796,7 +1796,7 @@ MibSModel::userFeasibleSolution(const double * solution, bool &userFeasible)
   int * lowerColInd = getLowerColInd();
   double * lpSolution = new double[getNumCols()];
   CoinFillN(lpSolution, getNumCols(), 0.0);
-  
+  MibSSolType solType;
   
 
   //bool bilevelbranch = MibSPar_->entry(MibSParams::isBilevelBranchProb);
@@ -1818,39 +1818,35 @@ MibSModel::userFeasibleSolution(const double * solution, bool &userFeasible)
   if(0)
     solver()->writeLp("userfeasible");
 
-  createBilevel(sol);
+  solType = createBilevel(sol);
 
-  for(i = 0; i < upperDim_; i++){
-      index = upperColInd[i];
-      lpSolution[index] = bS_->optUpperSolutionOrd_[i];
-      upperObj +=
-	  bS_->optUpperSolutionOrd_[i] * solver()->getObjCoefficients()[index];
-  }
-  for(i = 0; i < lowerDim_; i++){
-      index = lowerColInd[i];
-      lpSolution[index] = bS_->optLowerSolutionOrd_[i];
-      upperObj +=
-	  bS_->optLowerSolutionOrd_[i] * solver()->getObjCoefficients()[index];
+  if(solType != MibSNoSol){
+      for(i = 0; i < upperDim_; i++){
+	  index = upperColInd[i];
+	  lpSolution[index] = bS_->optUpperSolutionOrd_[i];
+	  upperObj +=
+	      bS_->optUpperSolutionOrd_[i] * solver()->getObjCoefficients()[index];
+      }
+      for(i = 0; i < lowerDim_; i++){
+	  index = lowerColInd[i];
+	  lpSolution[index] = bS_->optLowerSolutionOrd_[i];
+	  upperObj +=
+	      bS_->optLowerSolutionOrd_[i] * solver()->getObjCoefficients()[index];
+      }
   }
 
   userFeasible = false;
-  if(bS_->shouldPrune_){
+  if(solType == MibSRelaxationSol){
       userFeasible = true;
   }
 
-  if(!bS_->isProvenOptimal_){
-      userFeasible = false;
-  }
-  else if(userFeasible == true){
+  if(userFeasible == true){
       mibSol = new MibSSolution(getNumCols(),
 				lpSolution,
 				upperObj,
 				this);
   }
-  //else if(bS_->haveHeurSolCand_){
-  else if((((bS_->isLowerSolved_) || (bS_->isUBSolved_)) && (bS_->isProvenOptimal_)) ||
-	  ((!bS_->isLowerSolved_) && (bS_->tagInSeenLinkingPool_ ==
-				      MibSLinkingPoolTagLowerIsFeasible))){
+  else if(solType == MibSHeurSol){
       if(!bS_->isUBSolved_){
 	  isHeurSolution = checkUpperFeasibility(lpSolution);
       }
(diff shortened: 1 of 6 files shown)


src/MibSTreeNode.cpp
```diff
@@ -386,10 +386,12 @@ MibSTreeNode::process(bool isRoot, bool rampUp)
 	      tailOffTol = BlisPar->entry(BlisParams::tailOff);
 	  }
 	      
-	  if (ipSol) {         
-                // IP feasible
-                model->storeSolution(BlisSolutionTypeHeuristic, ipSol);
-		//model->storeSolution(BlisSolutionTypeRounding, ipSol);
+	  if (ipSol) {
```
tkralphs: Why is there a feasible solution stored at all if the node was infeasible?? This way of dealing with this still seems a bit clunky,

SaTahernejad: Yes, you are right. It does not seem reasonable to store a feasible solution (however, we know it is infeasible) when the node is infeasible. As I am checking MibS and Blis, the only way to not doing this is modifying Blis. The reason is that, after this line of Blis:

https://github.com/SaTahernejad/CHiPPS-BLIS/blob/master/Blis/src/BlisModel.cpp#L1792

The code goes to this function of MibS:

https://github.com/SaTahernejad/MibS/blob/d8b80f7d70b0cbd60f2824958b24bcb233a7b5f2/src/MibSModel.cpp#L1770

I can modify this function to set `userFeasible = true` and `mibSol = NULL` when the node is infeasible, but we will store a solution in Blis at this line:

https://github.com/SaTahernejad/CHiPPS-BLIS/blob/master/Blis/src/BlisModel.cpp#L1803

Hence, to never store any dummy solutions when the node is infeasible, Blis should be modified. Do you agree with me?

SaTahernejad: OK. I think if we set `userFeasible = false`, then we never store any solutions when the node is infeasible. Let me check it and come back to you with the result.

SaTahernejad: I committed new modifications. In this version, we do not store anything when the node is infeasible.

src/MibSModel.cpp
```diff
@@ -1836,12 +1836,12 @@ MibSModel::userFeasibleSolution(const double * solution, bool &userFeasible)
   userFeasible = false;
   if(bS_->shouldPrune_){
       userFeasible = true;
```
tkralphs: This logic is still a bit unclear. It's probably correct, but makes for code that is difficult to understand. Why is the indicator `shouldPrune_` used here to determine whether the solution is feasible? But then below is reversed if `isProvenOptimal_` is true? If nothing else, we could make this more transparent by doing something like
```
if (bS_->shouldPrune_ and bS_-> isProvenOptimal_){
    userFeasible = true;
    mibSol = new MibSSolution(getNumCols(),
                              lpSolution,
                              upperObj,
                              this);
}else{
   userFeasible=false;
}
```
or something like that to make the logic of when the solution is feasible or not more clear. What exactly does `isProvenOptimal_` indicate in general? That the lower level problem has an optimal solution? But this status is somehow combined with other information to determine whether the solution is actually bilevel feasible or not, which is what we need to know. The status that is actually returned externally should be something like just a solution status---feasible or infeasible. Does it make sense?

SaTahernejad: What you have suggested is correct and makes the code easier to understand. I think the reasonable way for setting the values of `shouldPrune_` and `isProvenOptimal_` should be as follows:
1. `shouldPrune_ = false` if and only if the node should not be pruned (in this case the value of `isProvenOptimal_` is not important).
2. `shouldPrune_ = true and isProvenOptimal_ = true` if the node should be pruned and that node contains at least one bilevel feasible solution. In this case, we prune the node after storing the optimal solution of that node.
3.  `shouldPrune_ = true and isProvenOptimal_ = false` if the node should be pruned and that node does not contain any bilevel feasible solutions.

In MibS, we set `shouldPrune_ = true` only when the node can be pruned, so the item 1 has been considered. However, the current version of MibS has not been implemented based on items 2 and 3. 
When we use the  the linking solution pool,  at a node in which the linking variables are fixed, we set `isProvenOptimal_ = false` in the case of finding those linking values in the pool and its status shows that the corresponding problem (UB) has been solved before. In this case, the node may include bilevel feasible solution, but we set `isProvenOptimal_ = false`. The logic is true because we have stored the optimal solution of this node before, but it makes the code difficult to be tracked. As you said, `isProvenOptimal_ = false` has a more general meaning than the situations in which the second-level problem or problem (UB) is infeasible. 

My suggestion is defining  a new member like `storeOptSol_` and set it to `true` whenever the node should be pruned and we want to store its optimal solution and set it to `false` in other cases. Do you agree with me?

tkralphs: In general, we probably shouldn't be be using private data members to pass information outside the class. Rather than adding a new data member, it would probably be better to add a method `shouldStoreSolution()` that returns `true` or `false`, but it's difficult to say because I don't have a good feel for how everything hangs together at the moment. It feels like the design has gotten a little off track. We had talked about doing a re-design at some point. Maybe we should do that :).

The main thing is not to have too many superfluous data members and to make sure the names communicate the meaning well. The member `isProvenOptimal` doesn't really seem to communicate what the data member is indicating (or perhaps what the data member is indicating can't really be understood outside the context of where it is set. `IsProvenOptimal` seems to be derived somehow from the status of the lower level solver when doing the bilevel feasibility check, but that should be translated later into the status we really care about, which is whether the solution to the relaxation is feasible or not, right? 

The logic in userFeasibleSolution should be pretty simple. A variable indicating whether the solution to the relaxation is feasible or not should be set and a solution should be stored if necessary. I would think that a data member or method indicating whether the solution to the relaxation is feasible or not makes sense and can be checked here to set set userFeasible. And then it seems like we just pass in a NULL pointer if we don't have a solution that needs to be stored. I guess we may want to distinguish between solutions obtained heuristically and solutions that are the solution to the relaxation, but that should be easy by checking whether userFeasible is true or not. Well, something like that. Does it make sense?

SaTahernejad: The idea of `shouldStoreSolution()` seems good.
I have a question about the `userFeasibleSolution` function. Consider the following case
1- the optimal solution of relaxation is not bilevel feasible.
2- the linking variables are fixed and we solve problem (UB) to find the optimal solution of node. 
3- the corresponding problem (UB) is feasible.
What should we set the value of `userFeasible` in this case? Also, should we set the optimal solution of relaxation as a heuristic solution or not?

tkralphs: You can look at the code to be sure, but logically, it would make sense that `userFeasible` is to indicate whether the solution to the relaxation is feasible (and this the node can be feasibly pruned). If we fix the linking variables and solve (UB), then if that is feasible, it is no different really than if we got it bya heuristic, right? This could in fact be considered a heuristic. All the underlying framework cares about is whether the solution was generated by solving the relaxation or not.

SaTahernejad:
I defined `MibSShouldStoreSol` type which can take these values
- `MibSShouldNotStoreAnything`
- `MibSShouldStoreRelaxationOptimal`
- `MibSShouldStoreHeurSol`
The function which checks the feasibility of the optimal solution of relaxation problem, returns one of the above values:
-`MibSShouldNotStoreAnything` in any of the below cases:
1- The optimal solution of relaxation problem is not bilevel feasible and no heuristic solution was not produced in the process of checking bilevel feasibility.
2- The linking variables are fixed and this linking value has been seen before.
- `MibSShouldStoreRelaxationOptimal` means that the optimal solution of relaxation is bilevel feasible.
- `MibSShouldStoreHeurSol` means that the optimal solution of relaxation is not bilevel feasible, but one heuristic solution was produced in the process of checking bilevel feasibility.

Then, we set `userFeasible = true` only if the returned value is `MibSShouldStoreRelaxationOptimal`, i.e., the optimal solution of relaxation is bilevel feasible.

In the current version of MibS, `isProvenOptimal = false` only if the corresponding second-level problem or problem (UB) is infeasible, this member does not show any information more than what it should.

PR #47, opened 2018-01-11
Some modifications in instanceStructure() function

SaTahernejad:
When user does not set the parameters of no-good cut (not the generalized one) and the type 1 of ICs, MibS turns them off.

### DIFF
--- a/src/MibSModel.cpp
+++ b/src/MibSModel.cpp
@@ -3172,7 +3172,8 @@ MibSModel::instanceStructure(const CoinPackedMatrix *newMatrix, const double* ro
 	std::cout << "Matrix G2 is positive." << std::endl;
     }
 
-    int paramValue(0), cutType(0);
+    int paramValue(0);
+    MibSIntersectionCutType cutType(MibSIntersectionCutTypeNotSet);
     //bool isHypercubeOn(false);
 
     bool turnOffOtherCuts(MibSPar_->entry(MibSParams::turnOffOtherCuts));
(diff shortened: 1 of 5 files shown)


src/MibSModel.cpp
```diff
@@ -3225,25 +3226,26 @@ MibSModel::instanceStructure(const CoinPackedMatrix *newMatrix, const double* ro
...
     //Param: "MibS_useInterSectionCut" (Type1)
-    if((turnOffOtherCuts == true) &&
-       (MibSPar_->entry(MibSParams::useIntersectionCut) == PARAM_NOTSET)){
-	MibSPar()->setEntry(MibSParams::useIntersectionCut, PARAM_OFF);
-    }
-    
     paramValue = MibSPar_->entry(MibSParams::useIntersectionCut);
+    cutType =  MibSPar_->entry(MibSParams::intersectionCutType);
+
+    if((paramValue == PARAM_ON) && (cutType == 1)){
```
tkralphs: It's good practice to use an `enum` rather than magic number here. We should have an `enum` of all `cutTypes`.

src/MibSModel.cpp
```diff
@@ -3225,25 +3226,26 @@ MibSModel::instanceStructure(const CoinPackedMatrix *newMatrix, const double* ro
...
+	if((isPureInteger_ == false) || (isLowerCoeffInt_ == false)){
+	    std::cout << "The intersection cut 1 does not work for this problem.  Automatically disabling this cut."
```
tkralphs: "is not valid" might be better than "does not work."

src/MibSModel.cpp
SaTahernejad: You are right. I take care of it.

src/MibSModel.cpp
SaTahernejad: I will change it.

src/MibSModel.cpp
SaTahernejad: Type 2 is the hypercube IC. Do you have any suggestions for the name of type 1? I am also thinking about it.

tkralphs: How about just `IntersectionCut` and `HypercubeIntersectionCut` or `IC` and `HypercubeIC`

tkralphs: Did you make this change yet? I guess not. I'd like to merge the PR and make a release.

SaTahernejad: In addition to IC and hypercubeIC, I had implemented two modifications of hypercube IC, but the result showed that hypercube IC is better than them. Hence, we have 4 type of IC in MibS, I will name them : IC, hypercubeIC, tenderIC and hybridIC. I will send the pull request in a few minutes.

PR #61, opened 2018-12-14
Showing help after typing mibs

SaTahernejad:


SaTahernejad:
It seems that MIBS_VERSION and MIBS_GIT_VERSION are not defined. I am not sure that how I can get them.

SaTahernejad:
I added the introduction into MibSMain.cpp. I think it can also be added to MibSModel.cpp. What is your suggestion?

tkralphs:
Can you move this to `master`? We should only be committing new things like this to master. We can then cherry pick or merge to stable branches.

tkralphs:
By the way, MIBS_VERSION and MIBS_GIT_VERSION should be defined by the configure script at build time in the autoconf header. You can check how this is done in other COIN projects.

tkralphs:
Can you also move this to master?

SaTahernejad:
> Can you also move this to master?

Sure. I will send all of them today.

PR #62, opened 2019-01-08
fix readme & ignore dependency folders

matbesancon:
Fetching dependencies was directly polluting the repo with other stuff.
Also updated the URL

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,19 @@ build*/*
 
 acinclude.m4
 autom4te*
+
+# Dependencies
+.subdirs
+Alps/
+Bcps/
+BuildTools/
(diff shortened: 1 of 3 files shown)


tkralphs:
Thanks for pointing out these issues. In the meantime, I implemented a different build script that doesn't pollute the MiBS source directory so I will update the README to reflect this latest change shortly.

tkralphs:
I've now updated the README in master. I would be happy if you would take a look and see if the new instructions make sense and work as specified. Sorry for just discarding this PR, but thanks for spurring me to update the README!

matbesancon:
no problem, the README is much clearer now

PR #72, opened 2019-08-13
Further changes for the cut paper: mostly about adding the parametric bound cut and debugging intersection cuts 

SaTahernejad:


SaTahernejad:
I cannot understand why there are conflicts. You have not changed any of the files that I have changed.

PR #75, opened 2019-12-05
For merge stochastic

SaTahernejad:


tkralphs:
@SaTahernejad Can you take a look at these conflicts?

SaTahernejad:
Sure. I will work on it today.

tkralphs:
The easiest is probably to rebase you branch onto current master.

SaTahernejad:
The errors are related to 2-dimensional vectors. I think I did not get such errors since I configured my code with the ADD_CXXFLAGS=-std=c++11 flag. I added this flag, first, to be able to use the random library.

tkralphs:
If you could avoid using any features that require C++11, that would be best. How much of a pain is it?

SaTahernejad:
As I said, the main reason that I used C++11 was using the random library. I should see how I can rewrite the code without using this library. 

Changing 2D vectors should not be difficult. I probably can substitute them with 2D arrays.

SaTahernejad:
Why should we avoid using C++11?

tkralphs:
A while back, we made the decision not to utilize any C++11 features in the COIN-OR code base (at least the projects in the Optimization Suite), basically for reasons of portability. These features are not supported by all compilers, especially older ones and, in any case, using them still requires additional compiler flags in many cases and I'm trying to keep the build and test uniform across all projects for easier maintenance. It's just avoids a lot of potential trouble not to use those modern features. Of course, if there is a killer feature that really makes a huge difference, we could decide to use it in MibS. MibS is not that widely used, so portability is not as much of an issue. It sounds like it would not be too difficult in this case to avoid C++11 so that would be better.

tkralphs:
@SaTahernejad just a ping to take a look at this.

tkralphs:
I just fixed a minor compilation issue that seems to be a clang thing. I don't see now what you mean about the use of C++11 and the random library. On Appveyor, everything is building and on Travis, we just seem have that one error that seems to be due to space needed between angle brackets. Could you take a look and remind what is going on from your perspective? I'd like to get this out as a release today if possible :).

SaTahernejad:
> I just fixed a minor compilation issue that seems to be a clang thing. I don't see now what you mean about the use of C++11 and the random library. On Appveyor, everything is building and on Travis, we just seem have that one error that seems to be due to space needed between angle brackets. Could you take a look and remind what is going on from your perspective? I'd like to get this out as a release today if possible :).

I am wondering that it did not complain about the random library. As I know, it needs C++11.
Actually, when I was working on the code, I could not build the code after adding the random library without C++11.

tkralphs:
OK, now the build is failing because of the use of `random` :). I guess C++11 support (or at least support for C++11 features in the stdlib) is standard on GCC 7 and the recent MSVC compiler, but not on GCC 5 or older clang :(. Can you do as we spoke about and put an `#ifdef` in to use std::random if C++11 support is there and not otherwise?

SaTahernejad:
We got that error finally :). Sure. I will debug it as we discussed.

src/MibSModel.cpp
```diff
@@ -37,6 +39,15 @@
...
 #include "OsiCbcSolverInterface.hpp"
+//#ifdef COIN_HAS_SYMPHONY
+#include "symphony.h"
+#include "SymConfig.h"
+#include "OsiSymSolverInterface.hpp"
+//#endif
+#ifdef COIN_HAS_CPLEX
+#include "cplex.h"
+#include "OsiCpxSolverInterface.hpp"
+#endif
```
tkralphs: Why are so many new headers included? Most of these header should probably be included from `MibSModel.hpp`, so that they are also automatically included into any file that includes `MibSModel.hpp`. But I would need to see why they are needed here in order to say.

src/MibSModel.cpp
```diff
@@ -55,6 +66,17 @@
...
+#if  COIN_HAS_MPI
+#include "AlpsKnowledgeBrokerMPI.h"
+#else
+#include "AlpsKnowledgeBrokerSerial.h"
+#endif
+
+//supporting c++11
+#if __cplusplus >= 201103L || (defined(_MSC_VER) && _MSC_VER >= 1900)
+#define HAS_C11
+#endif
```
tkralphs: This should actually be put into the configuration header in case we need to use it elsewhere. But you at least also need to wrap the `#include <random>` above in this `ifdef`, since it shouldn't be included unless C++11 support is present.

src/MibSModel.cpp
SaTahernejad: I see what you mean. I cannot remember any specific reasons. I will comment the ones which are not needed.

src/MibSModel.cpp
SaTahernejad: Oops! I forgot to put `#include <random>` in `ifdef`. I will put all of these in `MibSConfig.hpp`.

tkralphs: It's more complicated than that. `MibSConfig.hpp` is auto-generated. We would need to add a check to `configure.ac` that would then result in the symbol being defined. I can think about doing that. For now, what you are doing should be OK, but maybe put it in some more general header like `MibS.h` that is included everywhere.

SaTahernejad: Sure

tkralphs: What is the logic of that? I don't know what `MibSHelp.hpp` is supposed to be. It sounds like a header that would somehow be involved in the "help" functionality of MibS, i.e., information to help users with MibS. It only one thing in it, which itself looks like it should be somewhere else. So no, probably not there (and we should think about getting rid of `MibSHelp.hpp`). 

Looking at the various headers, I would say that `MibSConfig.hpp` is OK. It's not actually auto-generated itself, but it includes a header that's auto-generated and I thought it should remain prestine. However, we already have some stuff inserted in there that is specific to MibS and it is kind of the right place. So for now, you can just stick it in there.

src/MibSModel.cpp
```diff
@@ -12,7 +12,9 @@
 /* accompanying file for terms.                                              */
 /*===========================================================================*/
 
-#include "BlisModel.h"
+#include <random>
```
tkralphs: Hmm, you just introduced this header again without `ifdef`s... ?

src/MibSModel.cpp
SaTahernejad: Yes. We discussed before if we need `MibSHelp.hpp`. I quote it here :):

> Not sure we need a new file just for this. It should probably go into one of the other headers, like maybe MibSSolTypes.hpp :)? Actually, I didn't evven know that header file was there, looks like it's used for data structures to store solutions for heuristics, maybe.
> 
> Just for future reference, I guess that you meant something like MibSHelper.hpp? MibSHelp.hpp seems to indicate it's about the help message for the user or something (at least to me).
> 
> If we want to leave it in a separate header for different global types, I would call it something like MibSTypes.hpp.

So, as we discussed before, one option is renaming `MibSHelp.hpp` to either `MibSTypes.hpp` or `MibSHelper.hpp`. The other option is taking `MibSHelp.hpp` out and using one of the existing headers. What do you think? I prefer the first one as it may be useful in the future.

SaTahernejad:
> What was the logic of the changes in [f1ee621](https://github.com/coin-or/MibS/commit/f1ee621531718de4aa57bf3918cfa7a4af1569e1)? In general, any header needed by another header should just be included in that header, not in a source file where it happens to come before that header. There shouldn't typically be that many headers included directly in source files.

Some of the header files were included in both header and source files, so I deleted those ones. Do you agree with that? Now, I will put the included headers in `MibSModel.cpp` in `MibSModel.hpp`.

src/MibSModel.cpp
SaTahernejad: I intended to handle it in another commit :).

tkralphs:
Any progress :)?

SaTahernejad:
It seems the way of defining the `HAS_C11` macro does not work. I will work on it more.

tkralphs:
You just need to put these lines
```
#ifdef HAS_C11
#include <random>
#endif
```
below where `MibsConfig.hpp` is actually included. Where you put it, the symbol has not yet been defined. Also, if you can change it to `COIN_HAS_C11` for uniformity and to avoid potential name clashes, that would be helpful.

SaTahernejad:
Oops!

tkralphs:
OK, I think we are good now finally! Can I ask just one more thing, to document a little how to use the new parts of the code in the README?

tkralphs:
This is great! Could you also just add some brief documentation to the README?

SaTahernejad:
> This is great! Could you also just add some brief documentation to the README?

Sure. I am working on it.

README.md
```diff
@@ -257,11 +257,21 @@ e.g.,
...
-the header file `MibParam.h` for explanations. You can also find a detailed
+the header file `MibParams.hpp` for explanations. You can also find a detailed
 description of MibS
 [here](http://www.optimization-online.org/DB_FILE/2017/04/5977.pdf).
 Furthermore, to conduct the experiments illustrated in this report, see
-the `README` file in the directory `scripts`.     
+the `README` file in the directory `scripts`.
+
+`MibS` is also capable of solving two-stage mixed integer stochastic bilevel linear optimization problems. For an introduction to these problems, see this. To solve these problems, there are two ways:
+
```
SaTahernejad: Do we have any online references for stochastic problems?

README.md
```diff
@@ -257,11 +257,21 @@ e.g.,
...
+* You must provide an MPS file, a time file and a stoch file in the same way as the SMPS format. The second-stage objective coefficients should be defined at the end of the time file (see here). Then call `mibs` like this:
+```
+<build_or_install_dir>/bin/mibs -Alps_instance file.mps -MibS_auxiliaryTimFile file.tim -MibS_auxiliaryStoFile file.sto -MibS_stochasticityType stochasticWithoutSAA -MibS_isSMPSFormat 1 -MibS_isA2Random 1
+```
+The parameter `MibS_isA2Random` should be set to 0 in case the coefficients of the first-stage variables in the second-stage problem are not random variables.
+
+* You must provide an MPS file and an auxiliary information file in the same way as deterministic bilevel problems. The probability distributions of the random variables also should be specified by setting the values of corresponding parameters (MibS currently supports only the discrete uniform distribution). For a sample parameter file, see here.
```
SaTahernejad: Do you have any suggestions where I can put the sample parameter file?

tkralphs: Why not where the other parameter file is already? Just name it `mibs-stochastic.par.in` or something like that. By the way, we should clean up `mibs.par.in` a bit. I seem to have accidentally committed a modified version of it, for one, but it is also supposed to have paths that are auto-generated to be correct local paths. Instead, the paths are hard-coded. I guess I'll just accept this PR now and then we can work on cleaning up master.

PR #78, opened 2020-04-12
For merge stochastic

SaTahernejad:


SaTahernejad:
I modified README a little and added a parameter file for stochastic problems. I also modified mibs.par.in.

tkralphs:
All but the last two commits were already merged, so can you open a new PR? Or else, you can probably just commit them directly to master at this point. I'm not sure if I ever granted you write access, though.

PR #92, opened 2022-01-18
fix bugs in cut generator; clean up instance load

yuxies:
MibSCutGenerator:

1. generalNoGoodCut: condition check of SL-MILP, change from true to false;
2. benders interdiction cut: change a condition check and the initial value of valL;

MibSModel:

1. loadProblemData: remove an unused variable and redundant lines;
2. instanceStructure: add another condition check when assigning coefficient column signs; fix a bug when binary search rowIndex; fix typos and remove a redundant check when printing cut information.

### DIFF
--- a/src/MibSCutGenerator.cpp
+++ b/src/MibSCutGenerator.cpp
@@ -5631,15 +5631,21 @@ MibSCutGenerator::bendersInterdictionOneCut(BcpsConstraintPool &conPool, double 
       cutub += lObjCoeffs[i] * lSolution[i];
       valU = 0;
       valL = lObjCoeffs[i];
-      if(lSolution[i] > etol){
-          if (localModel_->colSignsG2_[i] == MibSModel::colSignPositive){ 
-             valU -= bigM;
-          } else {
-             valU += lObjCoeffs[i]*lSolution[i];
-          }
-      } else if (localModel_->colSignsG2_[i] == MibSModel::colSignPositive){
-         valL -= lObjCoeffs[i];
+      if (lSolution[i] > etol){
+         if (localModel_->colSignsG2_[i] == MibSModel::colSignNegative){ 
+            valU += lObjCoeffs[i]*lSolution[i];
+         } else {
+            valU -= bigM;
+         }
       }
+      //This case is in the paper, but it may not make sense in practice
+      //It only helps if the coefficient is negative, but then the solution
+      //lower-level solution could not be optimal anyway 
+      //else if (localModel_->colSignsG2_[i] == MibSModel::colSignPositive &&
+      //         lObjCoefficient < -etol){
+      //   valL -= lObjCoeffs[i];
+      //}
+
 
       if (valL){
          indexList.push_back(indexL);
@@ -6022,8 +6028,10 @@ MibSCutGenerator::generateConstraints(BcpsConstraintPool &conPool)
      }
      
      numCuts += feasibilityCuts(conPool) ? true : false;
-     
-     return (numCuts ? true : false);
+
+     //This return value indicates whether the relaxation needs to be re-solved
+     //and should always be false (see BlisTreeNode.cpp)
+     return (false);
 
   }else if(bS->isUpperIntegral_){
      if (localModel_->isInterdict_ && useBendersCut == PARAM_ON){
@@ -6065,7 +6073,9 @@ MibSCutGenerator::generateConstraints(BcpsConstraintPool &conPool)
         }
      }
 
-     return (numCuts ? true : false);
+     //This return value indicates whether the relaxation needs to be re-solved
+     //and should always be false (see BlisTreeNode.cpp)
+     return (false);
 
   }else if (useFractionalCuts){
      
@@ -6083,7 +6093,9 @@ MibSCutGenerator::generateConstraints(BcpsConstraintPool &conPool)
   
   //delete sol;
 
-  return numCuts ? true : false;
+  //This return value indicates whether the relaxation needs to be re-solved
+  //and should always be false (see BlisTreeNode.cpp)
+  return false;
 }
 
 //#############################################################################


src/MibSModel.cpp
```diff
@@ -852,10 +842,8 @@ MibSModel::loadProblemData(const CoinPackedMatrix& matrix,
       CoinDisjointCopyN(rowUB, numRows, conUB + auxULRows);
       
       /* Add interdiction budget row */
-      if (intCosts){
```
tkralphs, line 855: I lost the thread of why this if is not needed. Right now, it looks to me as though it is needed.

src/MibSModel.cpp
```diff
@@ -909,14 +897,12 @@ MibSModel::loadProblemData(const CoinPackedMatrix& matrix,
       
       /* Add interdiction budget row */
 
-      if (intCosts){
-         for(i = 0; i < auxULRows; i++){
```
tkralphs, line 913: I see why this is not needed because `auxULRows` should be zero when intCosts is 0, but the same logic doesn't apply to the above one.

src/MibSModel.cpp
```diff
@@ -3308,17 +3294,19 @@ MibSModel::instanceStructure(const CoinPackedMatrix *newMatrix,
...
-               colSignsA2_[i] = MibSModel::colSignPositive;
-            }else if (colSignsA2_[i] == MibSModel::colSignNegative){
-               colSignsA2_[i] = MibSModel::colSignInconsistent;
-            }
-         }else{
-            if (colSignsG2_[i] == MibSModel::colSignUnknown){
-               colSignsG2_[i] = MibSModel::colSignPositive;
-            }else if (colSignsG2_[i] == MibSModel::colSignNegative){
-               colSignsG2_[i] = MibSModel::colSignInconsistent;
+         } else if (!upperRow) {
```
tkralphs, line 3297: This seems correct, but to make the logic match more obviously, I guess we could instead do
```
         } else {
            if (!upperRow) {
               ...
            }
         }
```
It would be even more clear if we had `lowerRow` instead of `upperRow` and we swap the order so that we have
```
         if (mult * matElements[j] < -etol_){
            if (lowerRow){
               ...
            } else {
               ...
            }
         } else {
            if (lowerRow){
               ...
            }
         }
```
Does it make sense?

src/MibSCutGenerator.cpp
```diff
@@ -5630,9 +5630,9 @@ MibSCutGenerator::bendersInterdictionOneCut(BcpsConstraintPool &conPool, double
       indexL = lowerColInd[i];
       cutub += lObjCoeffs[i] * lSolution[i];
       valU = 0;
-      valL = lObjCoeffs[i];
+      valL = 0;
```
tkralphs, line 5633: This change definitely does not seem right to me. Can you explain? It doesn't match what is in the paper. It will effectively end up with `-d^2 y <= d^2 y^*` instead of `-d^2 y <= d^2 y^*`

tkralphs (changes requested):
I made some comments, could you address them and make the changes? You had also questioned some things in `instanceStructure()` that seemed like they should not be there. For example, around line 3175, `newRowSense` is being set, but nothing should really be set oin this function, it's just to analyze structure.

src/MibSModel.cpp
yuxies: Yes, since if not `intCosts`, then `auxULRows` is zero, and the fill function would not fill the bound. But I agree, keeping the condition can also remind about the dependency.

src/MibSModel.cpp
yuxies: We can also keep this if it makes the logic clear, like the above one.

src/MibSModel.cpp
yuxies: > This seems correct, but to make the logic match more obviously, I guess we could instead do
> 
> ```
>          } else {
>             if (!upperRow) {
>                ...
>             }
>          }
> ```
Yes makes sense.

> It would be even more clear if we had `lowerRow` instead of `upperRow` and we swap the order so that we have
> 

We currently don't have the variable `lowerRow`, and  `upperRow` was a result of a binary search. I feel that adding a new variable here just for a condition check may bring redundancy.

src/MibSCutGenerator.cpp
yuxies: Oh I think the variable is used to help load the coefficients of the term $\sum d^2_i y_i$ (at L5641), so it needs to be set to `0` at the beginning.

src/MibSModel.cpp
tkralphs: Yes, but if you use `lowerRow`, you don't need `upperRow` (they are complements of each other). The number of variables stays the same.  So can you make this change?

src/MibSModel.cpp
tkralphs: OK, yes, let's just keep it.

src/MibSModel.cpp
tkralphs: Yes, let's just keep it, why not.

src/MibSCutGenerator.cpp
tkralphs: It needs to be set to _something_ for sure, but 0 is wrong, as far as I can see. You didn't really explain _why_ you think it needs to be set to zero.

tkralphs: We should probably just delete the line 5641 below (and the enclosing `else if` block), since it only reduces the coefficients, which technically weakens the cut (it is a `<=`). I think my logic was that the corresponding variables would never be nonzero anyway, but it doesn't harm anything to give them nonzero coefficients.

src/MibSModel.cpp
yuxies: okay I can do that... but just want to say that at L3260 `upperRow` was probably designed to match `upperCol`, and the change of variable will also affect L3266, L3337, and L3339.

src/MibSCutGenerator.cpp
yuxies: I see. I forgot there is another term `d^2y` and only considered the conditioned summation. I will revert it back.

tkralphs: OK, these are only suggestions, change as you see fit.

src/MibSModel.cpp
```diff
@@ -3221,7 +3224,7 @@ MibSModel::instanceStructure(const CoinPackedMatrix *newMatrix,
    int nonZero (newMatrix->getNumElements());
    int counterStart, counterEnd;
    int rowIndex;
-   bool upperRow, upperCol;
+   bool lowerRow, upperCol;
```
tkralphs, line 3227: I was expecting you to also change `upperCol` to `lowerCol` for consistency :).

src/MibSModel.cpp
```diff
@@ -1029,7 +1028,7 @@ MibSModel::loadProblemData(const CoinPackedMatrix& matrix,
    setProblemType(); //determine the type of MIBLP
    //determine the list of first-stage variables participate in second-stage constraints
    setRequiredFixedList(newMatrix);
-   instanceStructure(newMatrix, conLB, conUB, rowSense);
+   instanceStructure(newMatrix, conLB, conUB, origRowSense_);
```
tkralphs, line 1031: We don't really need to pass data members as function arguments, we can just directly access the data member in the called function. Looking at this, all the arguments are really data that should be stored in the class, so this function really shouldn't have any arguments. If you think about it, it wouldn't make sense to analyze data that wasn't actually stored in the class somewhere. So let's just make it so this function has no arguments.

src/MibSModel.cpp
```diff
@@ -3107,7 +3106,7 @@ MibSModel::setRequiredFixedList(const CoinPackedMatrix *newMatrix)
 void                                                   
 MibSModel::instanceStructure(const CoinPackedMatrix *newMatrix,
                              const double* rowLB, const double* rowUB,
-                             const char *rowSense)
+                             const char *newRowSense)
```
tkralphs, line 3109: We can just change `newRowSense` to `rowSense` here (and throughout the function). The `new` part doesn't make sense out of context (all of this variable naming using `new` is not very intuitive and should ideally be changed, but we can tackle stuff like that later or just do it in version 2).

tkralphs (changes requested):

src/MibSModel.cpp
yuxies: I can remove rowSense, but `newMatrix`, `conLB`, and `conUB` are initialized locally in `loadProblemData`, so unless I add this to `initialize()`, these arguments are still needed for data access.

PR #97, opened 2022-03-05
Miscellaneous bug fixes for stable/1.2

yuxies:
Some problems found in checkBilevelFeasibility()

- L525: when entering the UB solving block, the corresponding parameter `isLinkVarsIntegral_` to `computeBestUBWhenLVarsInt` was missing; this reflects the mibs paper, where the parameter was also not included (for some unknown reason?) But I think it is still necessary...
- L652: this condition will check if `(x, \hat{y})`  is bilevel feasible and mark it as a heuristic solution; the tag here is changed to explicitly `LowerIsFeasible` because the old one includes `NotSet` case, and if linking solution pool is not used, the condition will always be satisfied. The case `LowerIsInfeasible` is already filtered at `!shouldPrune_`.
- L540 & L604 (not included in this (1st) commit as I am still thinking about a good way to rearrange):  these are two `TERM`s in the UB block when the time limit is reached. However, here, labeling the solution as `NoSol`  and jumping to termination directly ignores the heuristic solution `(x, \hat{y})` check at L652.

2nd commit summary (see details in the review comments):

- L652: moved the block `(x, \hat{y})` related to right after SL-MILP is solved, and marked `HeurSol` before entering the UB block; added sorting but not yet tested;
- L540 & L604: removed the statement because of the above change.
- running tests on the RAND_BILEVEL test set

### DIFF
--- a/src/MibSBilevel.cpp
+++ b/src/MibSBilevel.cpp
@@ -517,9 +517,7 @@ MibSBilevel::checkBilevelFeasiblity(bool isRoot)
 	    storeSol = MibSRelaxationSol;
 	}
 	else{
-		for (i = 0; i < lN; i++){
-			optLowerSolutionOrd_[i] = lowerSol[i];
-	    }
+	    memcpy(optLowerSolutionOrd_, lowerSol, sizeof(double) * lN);
 	    if(isUpperIntegral_){
 	        storeSol = MibSHeurSol;
 	    }
--- a/src/MibSBilevel.cpp
+++ b/src/MibSBilevel.cpp
@@ -522,7 +522,7 @@ MibSBilevel::checkBilevelFeasiblity(bool isRoot)
 	       (((branchPar == MibSBranchingStrategyLinking) &&
 		 (isIntegral_) && (isLinkVarsFixed_)) ||
 		((computeBestUBWhenXVarsInt == PARAM_ON) && (isUpperIntegral_)) ||
-		((computeBestUBWhenLVarsInt == PARAM_ON)) ||
+		((computeBestUBWhenLVarsInt == PARAM_ON)  && (isLinkVarsIntegral_)) ||
 		((computeBestUBWhenLVarsFixed == PARAM_ON) && (isLinkVarsFixed_)))){
 		if(UBSolver_){
 		    UBSolver_ = setUpUBModel(model_->getSolver(), objVal, false);
@@ -649,7 +649,7 @@ MibSBilevel::checkBilevelFeasiblity(bool isRoot)
 		    shouldPrune_ = true;
 		}	
 	    }
-	    else if ((tagInSeenLinkingPool_ != MibSLinkingPoolTagUBIsSolved) ||
+	    else if ((tagInSeenLinkingPool_ == MibSLinkingPoolTagLowerIsFeasible) ||
 		     ((!useLinkingSolutionPool) && (isUBSolved_))){
 		for (i = 0; i < lN; i++){
 		    index = lowerColInd[i];
--- a/src/MibSBilevel.cpp
+++ b/src/MibSBilevel.cpp
@@ -442,17 +442,26 @@ MibSBilevel::checkBilevelFeasiblity(bool isRoot)
 
 	    const double * values = lSolver->getColSolution();
 	    
+		// YX: round solution before saving or passing
+		for (i = 0; i < lN; i++){			 
+		    index = lowerColInd[i];
+		    if ((lSolver->isInteger(index)) &&
+			(((values[i] - floor(values[i])) < etol) ||
+			 ((ceil(values[i]) - values[i]) < etol))){
+			    lowerSol[i] = (double) floor(values[i] + 0.5);
+		    }else{
+			    lowerSol[i] = (double) values[i];
+		    }
+	    }
+
 	    if(useLinkingSolutionPool && isLinkVarsIntegral_){
-		std::copy(values, values + lN, shouldStoreValuesLowerSol.begin());
+		std::copy(lowerSol, lowerSol + lN, shouldStoreValuesLowerSol.begin());
 		
 		//step 12
 		//Adding x_L to set E  
 		addSolutionToSeenLinkingSolutionPool
 		    (MibSLinkingPoolTagLowerIsFeasible, shouldStoreValuesLowerSol, objVal_);
 		shouldStoreValuesLowerSol.clear();
-	    }
-	    else{
-		memcpy(lowerSol, values, sizeof(double) * lN);
 	    }
 	    
 	    MibSTreeNode * node = static_cast<MibSTreeNode *>(model_->activeNode_);
@@ -477,14 +486,9 @@ MibSBilevel::checkBilevelFeasiblity(bool isRoot)
     }
 
     //step 13
-    if(((!useLinkingSolutionPool || !isLinkVarsIntegral_) && (isProvenOptimal_)) ||
+    if(((!useLinkingSolutionPool || !isContainedInLinkingPool_) && (isProvenOptimal_)) ||
        ((tagInSeenLinkingPool_ == MibSLinkingPoolTagLowerIsFeasible) ||
 	(tagInSeenLinkingPool_ == MibSLinkingPoolTagUBIsSolved))){
-
-	OsiSolverInterface *UBSolver;
-	
-	//double *lowerSol = new double[lN];
-	//CoinFillN(lowerSol, lN, 0.0);
 
 	if(useLinkingSolutionPool && isLinkVarsIntegral_){
 	    //get optimal value  of (VF) from solution pool
@@ -513,7 +517,13 @@ MibSBilevel::checkBilevelFeasiblity(bool isRoot)
 	    storeSol = MibSRelaxationSol;
 	}
 	else{
-	    memcpy(optLowerSolutionOrd_, lowerSol, sizeof(double) * lN);
+		for (i = 0; i < lN; i++){			 
+		    pos = binarySearch(0, lN - 1, i, lowerColInd);
+			optLowerSolutionOrd_[pos] = lowerSol[i];
+	    }
+	    if(isUpperIntegral_){
+	        storeSol = MibSHeurSol;
+	    }
 	}
 	
 	if(!shouldPrune_){	
@@ -531,13 +541,12 @@ MibSBilevel::checkBilevelFeasiblity(bool isRoot)
 		    UBSolver_ =setUpUBModel(model_->getSolver(), objVal, true);
 		}
 
-		UBSolver = UBSolver_;
+		OsiSolverInterface * UBSolver = UBSolver_;
 
 		remainingTime = timeLimit - model_->broker_->subTreeTimer().getTime();
 
 		if(remainingTime <= etol){
 		    shouldPrune_ = true;
-		    storeSol = MibSNoSol;
 		    goto TERM_CHECKBILEVELFEAS;
 		}
 		
@@ -601,7 +610,6 @@ MibSBilevel::checkBilevelFeasiblity(bool isRoot)
 						       (dynamic_cast<OsiSymSolverInterface *>
 							(UBSolver)->getSymphonyEnvironment()))){
 		    shouldPrune_ = true;
-		    storeSol = MibSNoSol;
 		    goto TERM_CHECKBILEVELFEAS;
 		}
 		else if (UBSolver->isProvenOptimal()){
@@ -648,22 +656,6 @@ MibSBilevel::checkBilevelFeasiblity(bool isRoot)
 		    //isProvenOptimal_ = false;
 		    shouldPrune_ = true;
 		}	
-	    }
-	    else if ((tagInSeenLinkingPool_ == MibSLinkingPoolTagLowerIsFeasible) ||
-		     ((!useLinkingSolutionPool) && (isUBSolved_))){
-		for (i = 0; i < lN; i++){
-		    index = lowerColInd[i];
-		    if ((model_->solver()->isInteger(index)) &&
-			(((lowerSol[i] - floor(lowerSol[i])) < etol) ||
-			 ((ceil(lowerSol[i]) - lowerSol[i]) < etol))){
-			optLowerSolutionOrd_[i] = (double) floor(lowerSol[i] + 0.5);
-		    }else{
-			optLowerSolutionOrd_[i] = (double) lowerSol[i];
-		    }
-		}
-		if(isUpperIntegral_){
-		    storeSol = MibSHeurSol;
-		}
 	    }
 	}
     }
--- a/src/MibSBilevel.cpp
+++ b/src/MibSBilevel.cpp
@@ -283,7 +283,7 @@ MibSBilevel::checkBilevelFeasiblity(bool isRoot)
     double remainingTime(0.0), startTimeVF(0.0), startTimeUB(0.0);
     MibSSolType storeSol(MibSNoSol);
     int lN(model_->lowerDim_); // lower-level dimension
-    int uN(model_->upperDim_); // lower-level dimension
+    int uN(model_->upperDim_); // upper-level dimension
     int i(0), index(0), length(0), pos(0);
     int sizeFixedInd(model_->sizeFixedInd_);
     double etol(model_->etol_), objVal(0.0), lowerObj(0.0);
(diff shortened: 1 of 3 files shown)


tkralphs:
This looks good so far, thanks! I'll wait for that second commit. By the way, now that we have Github Actions working, it would be nice if `make test` did some actual testing. While you're working with the code, can you look into that a bit. At a minimum, just running a few instances of different classes and making sure the solutions are correct would be good. For example, you could look at how this is done in [Dip](https://github.com/coin-or/Dip/blob/master/Dip/test/Makefile.am). In the long run, some more robust unit testing would be good.

src/MibSBilevel.cpp
```diff
@@ -442,18 +442,27 @@ MibSBilevel::checkBilevelFeasiblity(bool isRoot)
 
 	    const double * values = lSolver->getColSolution();
 	    
+		// YX: round solution before saving or passing
```
yuxies, line 445: After solving SL-MILP, round the solution before feasibility check.

src/MibSBilevel.cpp
```diff
@@ -477,15 +486,10 @@ MibSBilevel::checkBilevelFeasiblity(bool isRoot)
     }
 
     //step 13
-    if(((!useLinkingSolutionPool || !isLinkVarsIntegral_) && (isProvenOptimal_)) ||
+    if(((!useLinkingSolutionPool || !isContainedInLinkingPool_) && (isProvenOptimal_)) ||
```
yuxies, line 489: Change the condition to include case `useLinkingSolutionPool && isLinkVarsIntegral_ && isProvenOptimal_ && TagNotSet`

- noticed that `tagInSeenLinkingPool` reads the history before feasibility check and is not updated in real-time.

src/MibSBilevel.cpp
```diff
@@ -477,15 +486,10 @@ MibSBilevel::checkBilevelFeasiblity(bool isRoot)
...
        ((tagInSeenLinkingPool_ == MibSLinkingPoolTagLowerIsFeasible) ||
 	(tagInSeenLinkingPool_ == MibSLinkingPoolTagUBIsSolved))){
 
-	OsiSolverInterface *UBSolver;
```
yuxies, line 484: Cleaned and moved to UB block.

src/MibSBilevel.cpp
```diff
@@ -513,7 +517,13 @@ MibSBilevel::checkBilevelFeasiblity(bool isRoot)
 	    storeSol = MibSRelaxationSol;
 	}
 	else{
-	    memcpy(optLowerSolutionOrd_, lowerSol, sizeof(double) * lN);
+		for (i = 0; i < lN; i++){
```
yuxies, line 520: Sort the solution when saving to `optLowerSolutionOrd_`; to match the saving LR solution in `createBilevel()` and the UB block.

- this needs more tests, and not sure yet how much difference it will make...

src/MibSBilevel.cpp
```diff
@@ -649,22 +657,6 @@ MibSBilevel::checkBilevelFeasiblity(bool isRoot)
 		    shouldPrune_ = true;
 		}	
 	    }
-	    else if ((tagInSeenLinkingPool_ == MibSLinkingPoolTagLowerIsFeasible) ||
```
yuxies: Moved above

- changed `model_->solver()` to `lSolver`, as the previous one refers to the LR solver;
- add rearrangement to match LR and UB solutions;
- if satisfied, saving as a ` heurstic ` solution is not affected by UB solved or not.

src/MibSBilevel.cpp
```diff
@@ -531,13 +541,12 @@ MibSBilevel::checkBilevelFeasiblity(bool isRoot)
...
 		}
 
-		UBSolver = UBSolver_;
+		OsiSolverInterface * UBSolver = UBSolver_;
 
 		remainingTime = timeLimit - model_->broker_->subTreeTimer().getTime();
 
 		if(remainingTime <= etol){
 		    shouldPrune_ = true;
-		    storeSol = MibSNoSol;
```
yuxies, line 540: Since `storeSol` was initialized as `MibSNoSol`, its type does not need to be "downgraded" in the case already marked as `MibSHeurSol`.

tkralphs:
Is this ready to go?

yuxies:
I'm running more tests for cuts affected by the change on the server... should have a conclusion tonight.

yuxies:
There was a memory error with the 2nd commit with message `free(): invalid next size (fast)` or invalid pointer; it happened when the tree was calling `removeDeadNodes()` recursively to `removeChild()` and destructing tree node. I guess there are some array copy/assignment issue but haven't found out why...

Also I saw (fractional) watermelon cut was able to find different objective solutions with a lower objective value for some instances.

yuxies:
Ran tests using dataset `dataIBLP-DEN` and 'MIBLP-XU', and compared the result with a build of master branch last year. 

- `dataIBLP-DEN`: using pure integer cut only and no heuristics; the objective values match for all solved instances, and the current version is able to solve 2 more instances.
- 'MIBLP-XU': using hypercube IC only (default option) and no heuristics; 8/100 instances obtained a different object value (and no better in terms of min), and 52/100 resulted in longer solution times. Since many functions changed in these two builds and this dataset involves continuous variables, not sure yet where the difference came from...

yuxies:
Following up the comments above, build version `684f6c3`,  ran tests on `MIBLP-XU, and compared the results between current commits vs. version `684f6c3`

- there is only 1 instance with a different objective value observed, which is actually a better min found in less solution time. So I think the commits in this pull request work fine.

src/MibSBilevel.cpp
```diff
@@ -442,18 +442,27 @@ MibSBilevel::checkBilevelFeasiblity(bool isRoot)
 
 	    const double * values = lSolver->getColSolution();
 	    
+		// YX: round solution before saving or passing
+		for (i = 0; i < lN; i++){			 
+		    index = lowerColInd[i];
+		    if ((lSolver->isInteger(index)) &&
```
yuxies, line 448: Question:

- Also wanted to confirm the variable rounding logic here: besides the tolerance requirement, do we round a variable when the LR solver says the variable should be an integer? or when the current (eg. SL-MILP) solver recognizes it? 
- The question came from the Master version, where in commit 4f2a5312a57696acecbc84ea715530695fe31711, the solver was changed from UBSolver to the LR solver when saving UB solutions.

tkralphs: > do we round a variable when the LR solver says the variable should be an integer? or when the current (eg. SL-MILP) solver recognizes it?

I don't really understand what you're asking. The rounding should be done when we have what is claimed to be an integer feasible solution by whatever solver gave us that solution, but some of the integer variables don't have exact integer values. I don't even know that we need to check tolerance at that point, since the solver has already claimed it as an integer solution. 

I didn't quite see what you are pointing to in 4f2a531 that is related. That commit doesn't look like it changed anything overall, other than allow for multiple blocks in the second-level problem..

src/MibSBilevel.cpp
```diff
@@ -517,8 +517,9 @@ MibSBilevel::checkBilevelFeasiblity(bool isRoot)
 	    storeSol = MibSRelaxationSol;
 	}
 	else{
-		for (i = 0; i < lN; i++){			 
-		    pos = binarySearch(0, lN - 1, i, lowerColInd);
+		for (i = 0; i < lN; i++){
+			index = lowerColInd[i];
+			pos = binarySearch(0, lN - 1, index, lowerColInd);
```
tkralphs: I don't see how this can be correct. `index` is in `lowerColInd` at position `i` by construction so `pos` will always be `i` here I think. This entirely change the logic here. But yeah, I guess I see how `i` was also not the right thing to have searched for in `lowerColInd`, since it is obviously not the index of the column, but an index of a position in lowerColInd (I guess you know what I mean).

The connection to the second commit above is beyond me, though, and I have no idea at all how this could fix a bug that resulted in a crash in `removeDeadNodes()` Could you explain a little more?

src/MibSBilevel.cpp
```diff
@@ -518,9 +518,7 @@ MibSBilevel::checkBilevelFeasiblity(bool isRoot)
 	}
 	else{
 		for (i = 0; i < lN; i++){
-			index = lowerColInd[i];
-			pos = binarySearch(0, lN - 1, index, lowerColInd);	
-			optLowerSolutionOrd_[pos] = lowerSol[i];
+			optLowerSolutionOrd_[i] = lowerSol[i];
```
tkralphs: OK, yes, I guess you realized that doing a binary search here doesn't make sense. But I'm still confused how such a fundamental bug could have persisted for so long. So the solutions MibS prints out have always been wrong?

src/MibSBilevel.cpp
```diff
@@ -517,9 +517,7 @@ MibSBilevel::checkBilevelFeasiblity(bool isRoot)
 	    storeSol = MibSRelaxationSol;
 	}
 	else{
-		for (i = 0; i < lN; i++){
-			optLowerSolutionOrd_[i] = lowerSol[i];
-	    }
+	    memcpy(optLowerSolutionOrd_, lowerSol, sizeof(double) * lN);
```
tkralphs, line 520: Hmm, OK, I'm confused...

tkralphs:
>Following up the comments above, build version https://github.com/coin-or/MibS/commit/684f6c3b9dc98018e9d0aa480f5b2ac2c3584b6e, ran tests on MIBLP-XU, and compared the results between current commits vs. version 684f6c3
> - there is only 1 instance with a different objective value observed, which is actually a better min found in less solution time. So I think the commits in this pull request work fine.

So there was a bug that was fixed then? What exactly is the reason why the one instance gets a different result. Can you figure out what commit exactly caused the solution value to change in that one instance. We should ideally really understand what happened here. Perhaps you have a guess.

tkralphs:
> using hypercube IC only (default option) and no heuristics; 8/100 instances obtained a different object value (and no better in terms of min), and 52/100 resulted in longer solution times. 

So master is buggy or what? What has changed that could make the difference? Can you try to figure out what the true optimal solutions are by doing some runs with and without cuts in both master and stable and see where there is agreement? Ideally, we would track down the bugs wherever there is disagreement to understand what is going on here.

I (think I) did some runs with MIBLP-XU both with and without cuts in stable/1.2 and I thought all the solutions were the same for the commit I was testing.

src/MibSBilevel.cpp
yuxies: > The rounding should be done when we have what is claimed to be an integer feasible solution by whatever solver gave us that solution

> I didn't quite see what you are pointing to in 4f2a5312a57696acecbc84ea715530695fe31711 that is related.

I was referring to the change at L819/870 and L833/884, where after the change the UB solutions are "judged" by the Relaxation `oSolver` but not the `UBSolver`. (Before this PR, the SL-MILP solution was also paired with the relaxation solver at [here](https://github.com/coin-or/MibS/blob/stable/1.2/src/MibSBilevel.cpp#L656).)

src/MibSBilevel.cpp
yuxies: I am still working on understanding how the indexing system works for `lowerSol` and its copies... so this last commit basically changed this single `memcpy` line back to before the pull request since it would provide the same output as before.

>  how this could fix a bug that resulted in a crash in `removeDeadNodes()`

The crash mentioned above was caused by the indexing error before the last fix, i.e. the unnecessary binary search using `i`: I guess some space with the wrong index was deleted twice when the tree manager cleans up after some iterations.

src/MibSBilevel.cpp
tkralphs: I see. I don't see exactly why that change was made, but an OsiSolverInterface object can store integrality information even when it is being use to solve the LP relaxation. I guess the solver object has the integrality information and it is being pulled from there for some reason, but the solution that is being rounded must have actually come from an MILP solver. I don't know without looking in more detail. It's also possible there's just a bug.

src/MibSBilevel.cpp
tkralphs: Hmm, OK, but why is this causing a problem now when it had been working fine...

yuxies: This still works fine and does not cause any problem right now! The problem was caused by the addition of `i` to `index` match that I added earlier (and now removed 😁 )

src/MibSBilevel.cpp
yuxies: Okay... And when setting up, the SL-MILP solvers and UB solvers "inherit" the integrality information from the LP relaxation solver. So if this integrality information in the LP solver does not change after obtaining solutions to the SL-MILP/UB problems, it can be reused.

Thank you 😀

tkralphs: Yes, what you said seems correct.

tkralphs:
What's the status here?

yuxies:
Ran branch-only experiments on MIBLP-XU and found some ??

- fractional branching by itself is even not sufficient to solve any instances in the set; usually after 1-2 iterations, the typical "cannot branching" error message shows; one example is the instance `bmilplib_10_6.out`, the default fractional branching + hypercube cut method cannot find any solution, but using only linking branching can find a solution after processing 10 nodes.
- (so that some of the differences may be attributed to the different branching methods?)
- linking branching method may have some issues with the cut generator setting since even though I turned on the hypercube cut and the output log confirmed that `Hypercube IC generator is on`, no cut was actually generated (according to the log), even for non-trivial instances that would find a solution after 1000+ nodes. So with or without the cut, the results were the same... 

So the investigation is still going on...

tkralphs:
Yes, you cannot use fractional branching without cuts. You have to be able to branch on variables that already have integer values because there is otherwise no way to continue the algorithm once an infeasible integer solution is produced. 

As for why no hypercube cuts are generated, I guess it should be easy to debug that. The generator is only called when the upper level solution is integer, so one way it would never be called is if the upper level solutions is never integer. I doubt that would happen, though. Once we have an integer infeasible solution, then we should always be able to cut it off with a hypercube cut. Can you verify whether the cut generator is being called and no cut is generated or it's just never being called?

yuxies:
The cut works fine! I forgot to change some parameters when running experiments...:cry: 

- so linking branching with vs. w/o hypercube IC: **3** instances that disagree on the objective value;
- linking branching vs. fractional branching, both with hypercube IC: **13** instances disagree; 

Still reading the logs and trying to find where the differences come from...

yuxies:
Regarding the last comment, I think the hypercube intersection cut removes one or more feasible nodes in the process, as the lower bound increases quickly after the cuts are added. Not sure if it is caused by numerical issues? 

More detailed running results of the test set can be found in the summary table in this [Google sheet.](https://docs.google.com/spreadsheets/d/1yIE3Afk_tQhXg9as2yWcPBfQYLw0798S/edit?usp=sharing&ouid=112426559758643619078&rtpof=true&sd=true)

tkralphs:
Any progress on this?

yuxies:
The above commit 5fc6339889fdbf86638eec8160efa32e51d97bec fixed the type of variable `useIncObjCut` in cut generator. 

- When not `turnOffOtherCuts`, because of a recent [condition change](https://github.com/coin-or/MibS/commit/225aebd8ae3e5fe7ff534b33b930e13ad80844f8#diff-849389bbcd814d3b187e1c2f3ab434dfe6af22e57069fa87a7c5d360bcbfcfccR3413) in MibSModel and var's ["int" type](https://github.com/coin-or/MibS/commit/f7c1011d578b42c4cee5236444fa7ef759790f3d#diff-a05333510ae67f0dd45fa67bdd1f56f3c94e6f2921d4e114c07f6ec9fd1f6dedR64) , the reading [here](https://github.com/coin-or/MibS/blob/stable/1.2/src/MibSCutGenerator.cpp#L5919) will force a `NOTSET` parameter to become `TURE`.

yuxies:
The difference was not caused by hypercube cuts, but by a solution verification condition check. 

- Since I moved the heuristic solution labeling of SL-MILP solution to the front before solving UB, when UB is solved but not optimal, the heuristic solution may be infeasible but the tag was not updated to `NoSol`. Then the feasibility check may be mistaken in some cases and pass a bilevel infeasible solution.

tkralphs (commented):
It's strange that this was a `bool`, while others were `int`. It seems to have been this way "since the beginning" for some reason. We should definitely be consistent. I also see at least a few other `bool` that perhaps should be `int` for consistency. 

However, switching to `int` seems to be patching over what is really another bug. No parameter should remain unset after initial parameter processing. The intent of the `NOTSET` indicator is to say that the user did not set the parameter and in that case, the parameter value should be set to some default, not left unset. In that way, the integer value should always be expected to be 0 or 1 and interpretable as a bool later on in the code. 

In the context here, we could really switch _all_ these variables to being bools, but then we should also switch to testing whether they are `TRUE`/`FALSE` rather than set to `PARAM_ON`. Perhaps it's better to leave it as is.

We could add asserts somewhere, though, to check that no parameter remains unset after processing. 

So in another words, it looks like my commit introduced a bug. Does it make sense?

tkralphs:
So are we good to go now, then?

yuxies:
> the parameter value should be set to some default, not left unset.

Now this cut is turned off in 72affa5b654084506bbe8971b73a1ab4afc01ee3 when not valid.

Did a few more tests today, and this PR should be ready!

PR #101, opened 2022-05-17
fix typo in function name: checkFeasibility

yuxies:
MibSBilevel.cpp: change function name; also add brackets to one line statement inside if;
MibSBilevel.hpp: change function name;


tkralphs (approved):
I'll merge this, but in the future, just try to make each commit atomic and to make pull requests only about one fix at a time. In this PR, you have a couple different unrelated things going on and the commit to fix the typo includes additional unrelated stuff.

PR #109, opened 2023-02-24
Add write and read functions for name based auxiliary files

yuxies:
Commits are copied from stable/1.2. 

Note: the master branch contains reading functions for stochastic case related files, and they are still using the index-based format.


tkralphs:
I think this is no longer relevant, so I'm closing. If it is , please feel free to re-open and comment.

PR #110, opened 2023-03-21
Add code to handle equality constraints

yuxies:
The equality constraints are handled at the beginning of `MibSModel::loadProblemData()`, after reading the `MPS` and `AUX` files, before further data loading. Then, `MibSModel` only recognizes the problem after parsing equality constraints. This is different from the original plan described in Slack as I was trying to place all the iterations in one block. 

Added variables: `inputLowerRowInd_`,  `inputLowerRowNum_` to track the original data from input files; they are necessary if writeAuxData is called.

Change in the output: printed information from `mps` functions will show the original row numbers, and those from `MibS` analyzing data will reflect the split row numbers.

### DIFF
--- a/src/MibSModel.cpp
+++ b/src/MibSModel.cpp
@@ -910,7 +910,7 @@ MibSModel::readProblemData()
    delete [] colType;
    delete [] varLB;
    delete [] varUB;
-   delete [] conLB; //YX: VERIFY POINTER
+   delete [] conLB;
    delete [] conUB;
    delete [] objCoef;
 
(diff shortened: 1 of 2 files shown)


tkralphs:
OK, I'm going to go ahead merge this. Everything seems OK with it and we can keep testing after it's merged.

src/MibSModel.cpp
```diff
@@ -287,7 +290,7 @@ MibSModel::readAuxiliaryData(int numCols, int numRows)
           std::cout << "MibS used " <<  fileName << " automatically.";
           std::cout << std::endl;
       }else{
-          fileName = mpsFile.erase(length - 3, 3);
+          fileName = mpsFile.substr(0, mpsFile.find("."));
```
yuxies, line 293: There is a "+1" missing here after find(). Without it, MibS won't be able to read .txt as aux files.

PR #111, opened 2023-03-28
add parser to handle empty LL/SL matrix

yuxies:
In MibSModel::loadProblemData(), if the lower level matrix is empty, add y_0 <= UB_(y_0) as a trivial row.


yuxies:
Temporary fix. Will need to check how to let SYMPHONY handle this.

PR #112, opened 2023-04-24
Read range constraints

yuxies:
Modify the equality constraint code block to handle other range constraints. Tested using linderoth.txt by adding 
`RANGES
    RHS       R0000004  20. ` 
or 
`RANGES
    RHS       R0000000  20. ` 
to the original problem. The `lpSolver.writeLp()` output matched the modifications. 

No change to writing functions, as they are either handled by `mps->writeMps` or already covered in the equality constraint commits.

### DIFF
--- a/src/MibSModel.cpp
+++ b/src/MibSModel.cpp
@@ -1013,7 +1013,7 @@ MibSModel::loadProblemData(const CoinPackedMatrix& matrix,
    CoinDisjointCopyN(lowerRowInd_, lowerRowNum_, inputLowerRowInd_);
 
    // YX: check if any equality/range constraint exists, modify row bounds, and 
-   //    append the splited cosntraint to the end with sense ">=";
+   //    append the splited constraint to the end with sense ">=";
    int numRows(inputNumRows);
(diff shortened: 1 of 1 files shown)


tkralphs:
This is really all that's needed? I am suspicious :). So no change is needed for writing out the aux file?

yuxies:
> So no change is needed for writing out the aux file?

I let the range constraints to be handled in the same way as the equality constraints (since equality constraints are range constraints eventually), where `mps-readMps` will parse the bounds and `loadProblemData` will convert one equality/range constraint into two (`L` + `G`) constraints. So MibS does not directly interact with the original preprocessed problem data. 

Then the only thing that needs special attention in writeAux is the lower-level number, which was kept track of by [having another variable storing the value here](https://github.com/coin-or/MibS/pull/112/commits/ad67c7e735eda8a3f7c30c3e35faeda48e01ced2#diff-849389bbcd814d3b187e1c2f3ab434dfe6af22e57069fa87a7c5d360bcbfcfccL1012).

tkralphs:
I see, OK. I guess I will just merge then and hope :).

PR #113, opened 2023-04-25
Read range constraints (continue): remove redundant condition check and fix bugs

yuxies:
Fixed some minor bugs. No (substantive) changes to read/write functions.

1. removed a redundant condition check for equality/range constraints (though I think C++ is using a short circuit logic so it may not be that redundant?);
2. added `break` to `findIndex()` so it terminates once the target is found;
3. added comments for reference.

### DIFF
--- a/src/MibSModel.cpp
+++ b/src/MibSModel.cpp
@@ -1022,9 +1022,7 @@ MibSModel::loadProblemData(const CoinPackedMatrix& matrix,
    std::vector<double> rowUBVec(rowUB, rowUB+inputNumRows);
    
    for(i = 0; i < inputNumRows; i++){
-      if((rowSense[i] == 'E') || 
-         ((rowLB[i] > -1*infinity) && (rowUB[i] < infinity))){
-         
+      if(((rowLB[i] > -1*infinity) && (rowUB[i] < infinity))){
          for(j = 0; j < inputLowerRowNum_; j++){
             if(inputLowerRowInd_[j] == i){
                lowerRowIndVec.push_back(numRows);
@@ -2211,13 +2209,13 @@ MibSModel::findIndex(int index, int size, int * indices)
   bool found(false);
 
   for(i = 0; i < size; i++){
-    if(indices[i] == index)
+    if(indices[i] == index){
       found = true;
+      break;
+    }
   }
 
   return found;
-
-
 }
 
 //#############################################################################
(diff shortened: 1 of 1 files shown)


yuxies:
The problem in the Slack chat was caused by unmatched column numbers in the .mps and .aux files. In debug mode, the error could be detected [here](https://github.com/coin-or/MibS/blob/stable/1.2/src/MibSModel.cpp#L1393). But I am not sure if we need to add a barricade to explicitly throw the error before any memory bug happens.

src/MibSModel.cpp
```diff
@@ -3642,7 +3640,7 @@ MibSModel::instanceStructure()
            case 'R':
              std::cout << "MibS cannot currently handle range constraints.";
              std::cout << std::endl;
-             abort();
+             abort(); // YX: handled in loadProblemData()
```
tkralphs, line 3643: I'm a little confused by why we're still aborting here if we can handle range constraints now.

src/MibSModel.cpp
```diff
@@ -1022,9 +1022,7 @@ MibSModel::loadProblemData(const CoinPackedMatrix& matrix,
    std::vector<double> rowUBVec(rowUB, rowUB+inputNumRows);
    
    for(i = 0; i < inputNumRows; i++){
-      if((rowSense[i] == 'E') || 
-         ((rowLB[i] > -1*infinity) && (rowUB[i] < infinity))){
-         
+      if(((rowLB[i] > -1*infinity) && (rowUB[i] < infinity))){
```
tkralphs, line 1025: I guess it's all the same, but why not `if (rowSense[i] == 'E' || rowSense =='R')`?

src/MibSModel.cpp
yuxies: If the constraints are read and handled correctly, then the code should not reach case "E" or "R" here. l thought these could be used as assertions so I didn't remove them.

src/MibSModel.cpp
yuxies: Okay, I can test and change that. I think they should be the same.

tkralphs: I was really asking the question "Why not?" I guess my argument for this is just that it makes the code asier to read if yu think it's correct.

src/MibSModel.cpp
tkralphs: OK, but at the very least, then the error message is not correct. We shouldn't tell the user that the problem is that MibS can't handle range constraints if it's just that there is a bug somewhere and this line should never be reached. Why not just an actual assert?

yuxies: So can I keep the `abort()` and then update the error message to "Something went wrong when converting equality(range) constraints"?

src/MibSModel.cpp
yuxies: Yes, I agree it would be easier to read. I couldn't recall why I used the other way... maybe I wasn't sure would `R` always be marked out when reading in the problem.

PR #115, opened 2023-06-01
add an option to write optimal value and solutions to a file

yuxies:
- If `MibS_writeInstanceName` followed by a string name is detected, the optimal value and solutions will be written to the given file name. 
- If there is no input following the parameter or the parameter is not used, results will be displayed through `std::out`. 

The output will look like this:
> ============================================
Optimal solution:
Cost = -2
First stage (upper level) variable values:
C0000001 = 1
C0000002 = 1
C0000003 = 1
Second stage (lower level) variable values:
C0000004 = 1
C0000005 = 1

### DIFF
--- a/src/MibSMain.cpp
+++ b/src/MibSMain.cpp
@@ -81,8 +81,15 @@ int main(int argc, char* argv[])
         AlpsKnowledgeBrokerSerial broker(argc, argv, model);
 #endif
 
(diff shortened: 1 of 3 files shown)


tkralphs:
OK, I made it so the solution always prints to the screen, which is I guess what we want. Looks good!

PR #118, opened 2023-07-14
Add error handler if invalid lowerObjCoeffs; fix a bug in writeAuxData

yuxies:
The added lines will check whether lower level (2nd stage) objective coefficients are given as expected in the input auxiliary file. If an empty/blank input is detected, `throw CoinError` to end the program.

Also fixed a bug in writeAuxiliaryData, where the `stringstream` content needs to be cleared before the next write.

### DIFF
--- a/src/MibSModel.cpp
+++ b/src/MibSModel.cpp
@@ -424,6 +424,9 @@ MibSModel::readAuxiliaryData(int numCols, int numRows)
 	 for(i = 0; i < lowerColNum; i++){
 	     data_stream >> cValue;
 	     data_stream >> dValue;
+         if(data_stream.peek() == EOF){
+            goto TERM_ENDREADAUXDATA;
+         }
 	     for(p = 0; p < numCols; ++p){
 		 if(columnName_[p] == cValue){
 		     pos = p;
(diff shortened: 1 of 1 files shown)


src/MibSModel.cpp
```diff
@@ -434,8 +437,10 @@ MibSModel::readAuxiliaryData(int numCols, int numRows)
 		 std::cout << cValue << " does not belong to the list of variables." << std::endl;
 	     }
 	     else{
-		 lowerObjCoeffs_[i] = dValue;
 		 lowerColInd_[i] = pos;
+         // YX: counter idx changed to match N at TERM
+         lowerObjCoeffs_[k] = dValue;
+         k++;
```
tkralphs, line 443: What on earth is going on with the indentation here :)? I guess that we should run the code through a reformatter at some point. But anyway, so I see now that the number of lower objective coefficients was already being counted in the case of the index-based format, we just weren't checking that we had the right number. Here, we don't need to count. `k` can't be different from `i`. All that can go wrong is that there are not two numbers on a given line, but only one, in which case what will fail is the conversion
```
data_stream >> dValue
```
because the next thing to be read won't be a double. So I guess that we need to check whether the reading succeeds or not, something like this: https://stackoverflow.com/a/18728791.

yuxies: > What on earth is going on with the indentation here :)?
The code format is using tabs but I only had space in my editor; the indentation looked ok so I didn't check the actual number of spaces needed...

> there are not two numbers on a given line, but only one, in which case what will fail is the conversion
This is actually handled at **L427**, where I was using the `EOF` flag check, and I think it would work the same as the `failbit` check.

> we don't need to count. k can't be different from i
Counting `k` is simply for throw error at the end of the function call at **L494**. I switched to `k` because (1) the index-based format is using `k`, and when there is an error in the input, `k` does not match the variable number; (2) then for both index-based and name-based formats, I can use `k != N` to determine whether the expected file format is received.

tkralphs:
I'm not sure why, but your commit here was attributed to `70344222+yuxies@users.noreply.github.com`. Can you check into that?

PR #124, opened 2023-10-10
skip the generation of Benders binary cut if the bigM problem is unbounded

yuxies:
Caused by unbounded \llv\ variables.

- If the bigM problem is determined unbounded or does not have an optimal solution, set the value of bigM to `-inf`.
- Add a boolean variable `isBigMIncObjErr_` to track this case, and turn off Benders binary cut once the error is detected.

Will run more tests.

### DIFF
--- a/src/MibSCutGenerator.cpp
+++ b/src/MibSCutGenerator.cpp
@@ -58,7 +58,6 @@ MibSCutGenerator::MibSCutGenerator(MibSModel *mibs)
   maximalCutCount_ = 0;
   numCalledBoundCut_ = -1;
   isBigMIncObjSet_ = false;
-  isBigMIncObjErr_ = false;
   bigMIncObj_ = 0.0;
   ImprovingDirectionICSolver_ = 0;
 }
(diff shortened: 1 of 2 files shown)


src/MibSCutGenerator.cpp
```diff
@@ -4835,8 +4834,17 @@ MibSCutGenerator::bendersBinaryCut(BcpsConstraintPool &conPool)
     }
     
     if(!isBigMIncObjSet_){
-       bigMIncObj_ = findBigMBendersBinaryCut();
-       isBigMIncObjSet_ = true;
+      bigMFound = findBigMBendersBinaryCut(bigMIncObj_);
```
tkralphs, line 4837: Just make this `isBigMIncObjSet_ = findBigMBendersBinaryCut(bigMIncObj_);` But please change `IncObj` to `BendBin` or just even make this `isBigMSet_`, since it's the only `bigM`.

src/MibSCutGenerator.cpp
```diff
@@ -4835,8 +4834,17 @@ MibSCutGenerator::bendersBinaryCut(BcpsConstraintPool &conPool)
...
+      // YX: skip if bigM is not found
+      if(bigMFound){
```
tkralphs, line 4839: Then we can eliminate the setting of `isBigMIncObjSet_` here.

src/MibSCutGenerator.cpp
```diff
@@ -4835,8 +4834,17 @@ MibSCutGenerator::bendersBinaryCut(BcpsConstraintPool &conPool)
...
+        isBigMIncObjSet_ = true;
```
tkralphs, line 4840: Then we can eliminate the setting of `isBigMIncObjSet_` here.

PR #126, opened 2023-10-14
Change the row bound rounding rule to tolerance based when set up the LL/SL problem

yuxies:
In the function `setUpModel()' in MibSBilevel:
- remove the rounding on the original row bounds `b^2`;
- add a tolerance check before rounding values `b^2 - A^2 x`.

### DIFF
--- a/src/MibSBilevel.cpp
+++ b/src/MibSBilevel.cpp
@@ -919,8 +919,8 @@ MibSBilevel::setUpModel(OsiSolverInterface * oSolver, bool newOsi,
       /** Set the row bounds **/
       for(i = 0; i < lRows; i++){
 	  index1 = lRowIndices[i];
-	  rowLb[i] = floor(origRowLb[index1] + 0.5);
-	  rowUb[i] = floor(origRowUb[index1] + 0.5);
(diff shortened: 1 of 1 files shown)


tkralphs (commented):
I don't think we even need `rowLb` and `rowUb` now. About all that happens is that we copy from one array to another, possibly negating the values. But then that new array is only used once after that and we could easily just have `origRowSense[index1] == 'G' ? origRowLb[i] : - origRowLb[i]` instead of `rowLb[i]`. What do you think?

yuxies:
We can, except for the initial setup. 

For `newOSI`, we need `rowLb` and `rowUb`  to store the rearranged \llv\ row bounds and send them to `loadProblem(*newMat, colLb, colUb, objCoeffs, rowLb, rowUb)`. After that, we can use additional condition checks, `origRowLb` and `origRowLb` when iterating over \llv\ rows. Some loops on \llv\ rows in `setUpModel()` can be further combined and cleaned up.

But since we have to use them once at the initial setup, we can also keep using them to make the code blocks easier(consistent?) to read.....

tkralphs:
OK, I see.

PR #127, opened 2023-11-02
Fix the node pruning logic when the time limit is reached

yuxies:
MibS feasibility check would label current node as should be pruned if timeout happens during solving the lower-level/second-stage problem. The terminating log will output the best lower bound as if the latest node is not yet processed.

Changes:
- Added a boolean variable in `MibSModel` as a flag once the preset time limit is reached during the feasibility check in `MibSBilevel`.
- In `MibSCutGenerator`, add the same exit flag when timeout during generating various intersection cuts.
- In `MibSTreeNode`, mark the most recent node exploration as "partially processed", so the termination process will keep the node knowledge instead of removing it.

### DIFF
--- a/src/MibSCutGenerator.cpp
+++ b/src/MibSCutGenerator.cpp
@@ -577,18 +577,11 @@ MibSCutGenerator::intersectionCuts(BcpsConstraintPool &conPool,
 		    double *lowerLevelSol = new double[lCols];
 		    CoinZeroN(uselessIneqs, lRows);
 		    CoinZeroN(lowerLevelSol, lCols);
-		    isTimeLimReached = false;
-		    if (!findLowerLevelSol(uselessIneqs, lowerLevelSol, sol, isTimeLimReached)){
+		    if (!findLowerLevelSol(uselessIneqs, lowerLevelSol, sol)){
                        delete [] uselessIneqs;
                        delete [] lowerLevelSol;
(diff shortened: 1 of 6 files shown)


tkralphs:
Is this still work-in-progress?

yuxies:
Verified the objective values for instances in IBLP-DEN, IBLP-FIS, and MIBLP-XU datasets.

src/MibSCutGenerator.cpp
```diff
@@ -583,11 +583,12 @@ MibSCutGenerator::intersectionCuts(BcpsConstraintPool &conPool,
                        delete [] lowerLevelSol;
                        goto TERM_INTERSECTIONCUT;
                     }
-		    if(isTimeLimReached == true){
```
tkralphs, line 586: So can we also remove the argument to `findLowerLevelSol` then, since it is not used?

yuxies: Yes, we can remove the `isTimeLimReached` from the argument list of `findLowerLevelSol()` and `findLowerLevelSolImprovingDirectionIC()`. I will update soon.

tkralphs:
The heuristics can be turned on, though, even if they're off by default. I'm going to merge this PR now to fix the immediate bug, but can you just do another PR for the heuristic timeouts? Are they difficult to fix?

PR #128, opened 2023-11-08
update linking related variable name and replace some binarySearch

yuxies:
- Variable name change: `sizeFixedInd` to `sizeLinkVars` to match a earlier name change.
- Update iterations for finding the linking variables to upper-level/first-stage only.
- In `MibSModel`, move the "check linking variables are integers" block from `readProblemData()` to `analyzeStructure()`; it also fix a bug on indices.

### DIFF
--- a/src/MibSBilevel.cpp
+++ b/src/MibSBilevel.cpp
@@ -160,14 +160,13 @@ MibSBilevel::createBilevel(CoinPackedVector* sol,
     }
   }
 
-  for(i = 0; i < N; i ++){
-      if(binarySearch(0, uN - 1, i, upperColInd) >= 0){
-	  if((varType[i] == MibSVarLinking) &&
-             (fabs(upper[i] - lower[i]) > etol)){
-	      isLinkVarsFixed_ = false;
-	      break;
-	  }
-      }
+  for(i = 0; i < uN; i++){
+    index = upperColInd[i];
+    if((varType[index] == MibSVarLinking) &&
+      (fabs(upper[index] - lower[index]) > etol)){
+        isLinkVarsFixed_ = false;
+        break;
+    }  
   }
   
   /* put the solution in order by integers first */
--- a/src/MibSModel.cpp
+++ b/src/MibSModel.cpp
@@ -138,7 +138,6 @@ MibSModel::initialize()
   upperRowNum_ = 0;
   origUpperRowNum_ =0;
   structRowNum_ = 0;
-  sizeLinkVars_ = 0;
   objAlignment_ = 0;
   counterVF_ = 0;
   counterUB_ = 0;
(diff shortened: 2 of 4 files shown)


src/MibSBilevel.cpp
```diff
@@ -160,14 +160,13 @@ MibSBilevel::createBilevel(CoinPackedVector* sol,
...
-	      break;
-	  }
-      }
+  for(i = 0; i < uN; i++){
+    index = upperColInd[i];
+    if((varType[index] == MibSVarLinking) &&
+      (fabs(upper[index] - lower[index]) > etol)){
+        isLinkVarsFixed_ = false;
+        break;
+    }
```
tkralphs, line 169: Yes, thank you, I had on my TODO list to get rid of these binary searches. Now that we have type for all variables rather than just listing the indices of linking variables, we don't need it.

src/MibSModel.cpp
```diff
@@ -138,7 +138,7 @@ MibSModel::initialize()
   upperRowNum_ = 0;
   origUpperRowNum_ =0;
   structRowNum_ = 0;
-  sizeFixedInd_ = 0;
+  sizeLinkVars_ = 0;
```
tkralphs: This doesn't even actually make sense anymore, since it was originally the size of the vector that stored the indices of the linking variables. Now, `varType_` always has length equal to the number of variables. I'll just push a change for this.

src/MibSModel.cpp
```diff
@@ -3479,37 +3467,40 @@ void
...
+          if(posRow >= 0){
+             varType_[index] = MibSVarLinking;
+             break;
+          }
+       }
+    }
+    
+    for(i = 0; i < lCols; i++){
+       index = lColIndices[i];
+       varType_[i] = MibSVarLower;
```
tkralphs, line 3503: Take a look at this change and just verify it.

tkralphs (commented):
I just pushed a commit to completely eliminate sizeLinkVars since it doesn't even make sense anymore. It was tracking the size of the vector of indices of the linking variable, but now `varType_` always has size equal to the total number of variables.

src/MibSModel.cpp
```diff
@@ -3493,14 +3494,15 @@ MibSModel::setVarTypes()
...
+             sizeLinkVars_ ++;
              break;
           }
        }
     }
     
     for(i = 0; i < lCols; i++){
        index = lColIndices[i];
-       varType_[i] = MibSVarLower;
+       varType_[index] = MibSVarLower;
```
tkralphs, line 3505: Thanks for spotting this!

tkralphs:
Is this still a draft?

yuxies:
Update：

- Reorganized `analyzeStructure(}`: use row or column indices to replace some binary searches.

PR #130, opened 2023-11-29
Fix a bug when setting up the improving direction intersection cut

yuxies:
When IDIC parameter is not set, add a condition in the check to make sure that the cut cannot be turned on if some of the lower objective coefficients are not integers.


tkralphs:
Ah, so there was a check for the lower-level objective being integer, but it was only applied when the user actually set the parameter for the cut? If it was turned on by default, it was not checked?

yuxies:
Yes

PR #135, opened 2024-02-01
Change default ISIC message to (Type I)

yuxies:
In MibSModel::printProblemInfo(), change the default Improving Solution Intersection Cut message to Type I cut.

### DIFF
--- a/src/MibSModel.cpp
+++ b/src/MibSModel.cpp
@@ -4149,10 +4149,10 @@ MibSModel::printProblemInfo(){
     
     if (MibSPar_->entry(MibSParams::useImprovingSolutionIC) == PARAM_ON){
        if (MibSPar_->entry(MibSParams::bilevelFreeSetTypeISIC) ==
-           MibSBilevelFreeSetTypeISICWithLLOptSol){
-          std::cout << "Improving solution intersection cut generator (Type I) is on." << std::endl;
-       }else{
+           MibSBilevelFreeSetTypeISICWithNewLLSol){
           std::cout << "Improving solution intersection cut generator (Type II) is on." << std::endl;
+       }else{
+          std::cout << "Improving solution intersection cut generator (Type I) is on." << std::endl;
        }              
     }
 
--- a/src/MibSModel.cpp
+++ b/src/MibSModel.cpp
@@ -4149,10 +4149,10 @@ MibSModel::printProblemInfo(){
     
     if (MibSPar_->entry(MibSParams::useImprovingSolutionIC) == PARAM_ON){
        if (MibSPar_->entry(MibSParams::bilevelFreeSetTypeISIC) ==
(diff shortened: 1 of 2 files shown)


tkralphs:
This seems correct to me the way it currently is so I'm going to close this. You can reopen if you disagree.

yuxies:
? But in output info, MibS would print "Type II" even though "Type I" was called. The default intersection cut was set to Type I, not Type II.

tkralphs:
I'm lost. If `bilevelFreeSetTypeISIC` is set to `MibSBilevelFreeSetTypeISICWithLLOptSol`, that is a Type I cut and it prints out Type I in that case. Otherwise, it prints out Type II. It matches. I don't know what this has to do with the default. It's just printing out what the parameter is set to, whether it's set to that because it's the default or because the user set it manually, it's still the parameter value. I guess I'm missing something, but I still don't get what's going on here.

yuxies:
If user did not set which ISIC, `bilevelFreeSetTypeISIC` would be `MibSBilevelFreeSetTypeISICNotSet`. Then it will print out Type II while applying Type I.

tkralphs:
Argh, OK, you had explained this to me before, but the problem is that this is very difficult to see from looking at the code and the fix should make it more transparent. We have a standard mechanism for indicating a parameter's value is not set, which is to set it's value to `PARAM_NOT_SET` (`-1`). Here, we are setting the parameter to value `-1`, but the first issue is that we are using a different symbol for that, which is not very transparent. We should be `PARAM_NOT_SET` rather than `MibSBilevelFreeSetTypeISICNotSet`.

But on top of that, the reason to allow parameters not to be set in the first place is because we want to know that they are not set in order to automatically set them to possibly different values, depending on the instance. In this case, I guess that we are always using the same default value no matter what, so there is no reason not to just set that as the default instead of setting it to `MibSBilevelFreeSetTypeISICNotSet`. Then everything is fine.

yuxies:
Oh I thought we agreed that something went wrong... i should have listed enough details in the front.

> just set the default value to whatever we want it to be right away.

Then instead of correcting the condition check, when we see `MibSBilevelFreeSetTypeISICNotSet`, we would change this to `MibSBilevelFreeSetTypeISICWithLLOptSol`, i.e. Type I cut, in the adjust parameter function in MibSModel. Do you think it is a better fix?

tkralphs:
What I mean was just to set the parameter value to a default in `MibSParams` like many other parameters are already set. The only reason not to just set a default, but rather to leave the parameter's value unset is so that we can detect whether or not the user set the parameter's value manually or not. If there is a default value, we can't know whether the user manually set the parameter's value to that same default (in which case we should leave it as is) or we can change from that default after analyzing the problem structure. You don't need to change any code anywhere else except to eliminate `MibSBilevelFreeSetTypeISICNotSet`, since the parameter's value will always be set.

Issue #134, opened 2024-02-01
C++ Hello World - Segmentation Fault


hlefebvr:
Hello,

I am trying to build a minimal example for using MibS with C++. 

# Context

I am trying to model the following problem using the C++ API:
```
 min  -x − 7y
 s.t.   -3x + 2y ≤ 12
             x + 2y ≤ 20
             x ≤ 10
             y ∈ arg min {z : 2x - z ≤ 7,
                              -2x + 4z ≤ 16,
                              z ≤ 5}
```

# Setting 

- MibS 1.2 as built using coinbrew and linked with Cplex - though the problem I am facing seems indifferent to CPLEX. 
- Ubuntu 22.04 LTS
- CMake as build system
- C++17

# My Attempt

So far, I have tried with the following code. 
```cpp

    std::unique_ptr<OsiSolverInterface> solver(new OsiCpxSolverInterface());

    MibSModel model;
    model.setSolver(solver.get());

   // AUXILIARY DATA
    std::vector<int> upper_level_variables_indices = { 0,  };
    std::vector<int> lower_level_variables_indices = { 1,  };
    std::vector<int> upper_level_constraints_indices = { 0, 1,  };
    std::vector<int> lower_level_constraints_indices = { 2, 3,  };
    std::vector<double> lower_level_objective_coefficients = { 1,  };

    model.loadAuxiliaryData(
            (int) lower_level_variables_indices.size(),
            (int) lower_level_constraints_indices.size(),
            lower_level_variables_indices.data(),
            lower_level_constraints_indices.data(),
            1.,
            lower_level_objective_coefficients.data(),
            (int) upper_level_variables_indices.size(),
            (int) upper_level_constraints_indices.size(),
            upper_level_variables_indices.data(),
            upper_level_constraints_indices.data(),
            0,
            nullptr,
            0,
            nullptr,
            nullptr,
            nullptr
    );

   // PROBLEM DATA

    std::vector<double> variable_lower_bounds = { 0, 0,  };
    std::vector<double> variable_upper_bounds = { 10, 5,  };
    std::vector<char> variable_types = { 'I', 'I',  };
    std::vector<double> constraint_lower_bounds = { -1e+20, -1e+20, -1e+20, -1e+20,  };
    std::vector<double> constraint_upper_bounds = { 12, 20, 7, 16,  };
    std::vector<char> constraint_types = { 'L', 'L', 'L', 'L',  };
    std::vector<double> objective = { -1, -7,  };

[...]

PR #141, opened 2024-09-03
Check input matrix orientation when loading problem data

yuxies:
To fix the bug in #134: since `loadProblemData()` expects a column-oriented data matrix, we consider adding a check to the beginning of the function and creating a reversed copy if a row-oriented matrix is detected.

### DIFF
--- a/src/MibSModel.cpp
+++ b/src/MibSModel.cpp
@@ -1083,7 +1083,8 @@ MibSModel::loadProblemData(const CoinPackedMatrix& matrix,
    int problemType(MibSPar_->entry(MibSParams::bilevelProblemType));
 
    int i(0), j(0), k(0), start(0), end(0), beg(0);
-
+   
+   bool matType = matrix.isColOrdered(); // YX: check input matrix orientation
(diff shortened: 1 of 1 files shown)


tkralphs:
Copying and changing the orientation are expensive operations, do we really need to make the copy? Don't we won the matrix? Also, I wonder if we should just throw an error instead, forcing the user to give us the right thing? Otherwise, they may not even realize that they are constructing the matrix in the wrong orientation to begin with.

yuxies:
We have to make a copy to change the col/row order; the function `loadProblemData()` does not own the matrix since it is passed in with the `const` keyword. 

>  I wonder if we should just throw an error instead, forcing the user to give us the right thing? Otherwise, they may not even realize that they are constructing the matrix in the wrong orientation to begin with.

I was also thinking about this strategy to avoid changing any part of the function itself. It is natural for the user to create a row matrix when they input data by adding rows to the matrix. So if we throw an error, we can inform the user that they are creating a row matrix, while the function defaults to using a column matrix.

tkralphs:
I guess that throwing an error is better.

PR #145, opened 2024-10-29
Adding parameters to allow more fine-grained control of cut generation

tkralphs:


src/MibSBilevel.cpp
```diff
@@ -420,6 +427,7 @@ MibSBilevel::checkBilevelFeasibility(bool isRoot)
 	}else{
 	  startTimeVF = model_->broker_->subTreeTimer().getTime();
 	  lSolver->branchAndBound();
+          lSolver->writeLp("water");
```
febattista, line 430: I think this is a typo, shouldn't this be commented out?

src/MibSCutGenerator.cpp
```diff
@@ -6012,14 +6012,25 @@ MibSCutGenerator::generateConstraints(BcpsConstraintPool &conPool)
...
-     if (useFractionalCuts && useImprovingDirectionIC == PARAM_ON){
+     if (useImprovingDirectionIC == PARAM_ON &&
+         ((haveSecondLevelSol &&
+           relaxedObjVal > localModel_->bS_->objVal_ + localModel_->etol_) ||
+          (localModel_->MibSPar_->entry(MibSParams::bilevelFreeSetTypeISIC) ==
+           MibSBilevelFreeSetTypeISICWithNewLLSol &&
+           (ISICGenStrategy == MibSIDICGenStrategyLInt ||
+            ISICGenStrategy == MibSIDICGenStrategyAlways ||
+            (ISICGenStrategy == MibSIDICGenStrategyAlwaysRoot &&
+             localModel_->activeNode_->getDepth() == 0))))){
```
febattista, line 6033: I can't make sense out of this straightforwardly. Here the code should decide whether to undertake the generation of an IDIC, but is checking for the bilevel free set type for ISIC. Moreover, I might be wrong but I guess here `ISICGenStrategy` should be replaced with `IDICGenStrategy`.

src/MibSCutGenerator.cpp
```diff
@@ -6044,17 +6055,43 @@ MibSCutGenerator::generateConstraints(BcpsConstraintPool &conPool)
...
+     if (useImprovingSolutionIC == PARAM_ON &&
+         ((haveSecondLevelSol &&
+           relaxedObjVal > localModel_->bS_->objVal_ + localModel_->etol_) ||
+          (localModel_->MibSPar_->entry(MibSParams::bilevelFreeSetTypeISIC) ==
+           MibSBilevelFreeSetTypeISICWithNewLLSol &&
+           (ISICGenStrategy == MibSIDICGenStrategyYInt ||
+            ISICGenStrategy == MibSIDICGenStrategyAlways ||
+            (ISICGenStrategy == MibSIDICGenStrategyAlwaysRoot &&
+             localModel_->activeNode_->getDepth() == 0))))){
+        cutType = MibSIntersectionCutImprovingSolution;
```
febattista, line 6076: Similarly as above, I guess `MibSIDICGenStrategyYInt`, `MibSIDICGenStrategyAlways` and `MibSIDICGenStrategyAlwaysRoot` should be replaced with `MibSISICGenStrategyYInt`, `MibSISICGenStrategyAlways` and `MibSISICGenStrategyAlwaysRoot`, respectively.

